Issue #11, opened 2016-11-23
Questions and features/bugs


lludol:
Hello,

Thank you for gitmoji, it's fun and usefull to have an emoji norm to commit message!

1) Default behaviour
```shell
> gitmoji
>
```

Nothing happens when we run the cli without parameter.
I read your code and it's because you run a command only if you have a parameter.

I think it's better to run ```gitmoji -c``` by default.

2) Git add

```shell
> gitmoji -c
? Choose a gitmoji: ✏️  - Fixing typos.
? Enter the commit title: test
? Enter the commit message: toto
? Issue / PR reference #: 
? Signed commit: No

ERROR: 
```

Can you check if we added some files before running gitmoji ?
Or can you just display a real error message (for example: "Error: you have to add some files first") ?

3) Commitizen

Do you know the commitizen project ?
They also have a cli : https://github.com/commitizen/cz-cli

At the beginning, I wanted to create a gitmoji config for commitizen but I can't find a repository which only export the ```gitmojis.json```.

4) Contributing

Can we create PRs to fix bugs / add features or you prefer that we add issues ?

carloscuesta:
Thanks for filling an issue:

1. Yes I know nothing happens when you run gitmoji, I considered adding -c as a default option and this will be done, it's better to have a default command than nothing.

2. Good point, at first instance I thought that git will do this and print an error, will take a look at that. By default the -c option adds the files before commiting.

3. Never heard of commitizen but it's a really cool project, I'm sure we can create a custom adapter for the people that is using commitizen.

4. My bad! I didn't add CONTRIBUTING guidelines, will fix this as soon as possible, feel free to open issues and after that sending a pull request refering that issue.

Thanks @lludol 😄

carloscuesta:
:memo: [Contributing docs added](https://github.com/carloscuesta/gitmoji-cli/tree/master/.github)

PR #16, opened 2016-11-24
Fix default behaviour and display help if invalid args

lludol:
## Description

I have refactored the code to replace multiple "if" by an Object with arrow function.
When we run the cli without args, it will run the commit command.
If args are invalid, the help will be displayed.

Issue: #11 

## Tests

- [x] All tests passed.


carloscuesta:
That's cool @lludol thanks for that pull request! 👍

Issue #18, opened 2016-11-26
Create a commitizen version for gitmoji


carloscuesta:

thibmaek:
I might try to get this one PR'd tomorrow if I find the time ;)

Landish:
Would about this https://github.com/Landish/cz-gitmoji ? Would you mind?

carloscuesta:
Cool! @Landish I'm goint to add as a reference on the docs 🎉

PR #23, opened 2016-12-08
:sparkles: add commitizen prompter

zcei:
## Description

Adds a commitizen adapter as requested in #18.
To achieve this I moved the axios API client to an own file & pulled the message string generation out of the function that really commits.

Tests are still running, but there are none for commitizen prompters.
Currently working on a stub/mock, so that we can use some sort of integration tests, where we expect a set of questions and return a then function that gets predefined answers passed in. Then we can assert the value of the `commit` callback. Probably smth like [here](https://github.com/commitizen/cz-jira-smart-commit/blob/master/index.test.js)
Just wanted to share the work beforehand 🙂 

Issue: #18 

## Tests

- [x] All tests passed.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -2,9 +2,9 @@
 'use strict';
 
 const meow = require('meow');
-const axios = require('axios');
 const updateNotifier = require('update-notifier');
-const GitmojiCli = require('./src/gitmoji.js');
+const GitmojiCli = require('./src/gitmoji');
+const gitmojiApiClient = require('./src/api-client');
 const pkg = require('./package.json');
 
 updateNotifier({pkg}).notify();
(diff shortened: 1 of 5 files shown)


carloscuesta:
Thanks! @zcei 

I'm thinking about setting up the commitizen version to a different github repository and link it to the README 🤔. 

In my opinion would be easier to find the commitizen version for gitmoji if there is a specific repo for that.

Would you like to create something like gitmoji-commitizen repository?

zcei:
I thought about the same, but as you already had an issue for this, I wanted to PR this one 😛 

For me it would make most sense, if the gitmojis json itself would be a repo, so that the website, the cli, and the commitizen version can all depend on it, instead of doing manual HTTP calls.

If you agree, I'd love to see a `gitmoji` organization instead of having the repos being spread over multiple user names. What do you think?

carloscuesta:
Probably an organization is too much "closed" for people to create new projects and repositories against gitmoji, I don't think that we really need an organization 🤔

The `gitmojis.json` now are being served from the "main repository" of the project, and for me that makes sense and it's good 😸

zcei:
Ok sure: your project - your structure 🙂 

Would you still accept the PR if I remove the commitizen part? Then I could at least just depend on your cli without reinventing the wheel (aka the questions & formatting).
That would help keeping the module up-to-date without unnecessary maintenance.

carloscuesta:
Resolved at #18 🎉

PR #42, opened 2017-04-15
Title Template Support :sparkles:

avaly:
## Description

Adds a new configuration option to customize the commit title template

## Tests

- [x] All tests passed.

### DIFF
--- a/src/gitmoji.js
+++ b/src/gitmoji.js
@@ -19,6 +19,9 @@ class GitmojiCli {
 		if (config.get('autoadd') === undefined) {
 			config.set('autoadd', true);
 		}
+		if (config.get('titleTemplate') === undefined) {
+			config.set('titleTemplate', '{gitmoji} {title}');
+		}
 	}
 
 	config() {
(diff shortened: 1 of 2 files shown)


carloscuesta:
Hey @avaly 😄! 

Thanks for sending a Pull Request. Honestly, I don't think we need to template the commit title. Gitmojis should be used at the start of a commit to identify the purpose of a commit at first without reading the whole commit.

Before implementing any feature, feel free to create an issue first 💪

avaly:
Hi @carloscuesta! Sorry to hear you don't feel this is worth merging. I actually prefer to use my emoji at the end of the commit title. I'll use my fork instead.

carloscuesta:
Feel free @avaly 😄

angelikao:
Sad that the idea was rejected because in my current project it is required by the client to put issueId first and then we could put gitmoji before the describing commit message, so for example 'ISS-102 💯 Adding some description'.

carloscuesta:
Then probably gitmoji it’s not the convention you need to use for that particular project! However you can always try to negotiate the commits convention and explain the benefits @angelikao

PR #53, opened 2017-06-15
✨ Add possibility to remove hook

thibmaek:
## Description

Introduces the new command `--remove|-r` to unlink a previously set hook. Avoids having to dive into the .git dir in a terminal and remove it there. I love Gitmoji, but sometimes needed to remove it for projects where other people didn't want to use it.

It uses `fs#unlink` which is an async method that takes the path and a completion callback. Very much like `fs#writeFile`.

## Tests

- [x] All tests passed.

### DIFF
--- a/src/gitmoji.js
+++ b/src/gitmoji.js
@@ -74,11 +74,11 @@ class GitmojiCli {
 			fileContents
 		};
 	}
-	
-	unhook() {
+
+	remove() {
 		const hookFile = 'prepare-commit-msg';
 		const path = `${process.env.PWD}/.git/hooks/${hookFile}`;
-		
+
 		if (this._isAGitRepo('.git')) {
 			fs.unlink(path, err => {
 				if (err) {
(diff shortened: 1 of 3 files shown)


src/gitmoji.js
```diff
@@ -75,6 +75,24 @@ class GitmojiCli {
 		};
 	}
 
+	remove() {
+		const hookFile = 'prepare-commit-msg';
```
carloscuesta: Can you delete `hookFile` and `path` constants from `remove()` and init `methods()` and create a global `commitHookPath` constant outside the class. Code duplication is dangerous! ⚠️

src/gitmoji.js
```diff
@@ -75,6 +75,24 @@ class GitmojiCli {
...
+		const path = `${process.env.PWD}/.git/hooks/${hookFile}`;
```
carloscuesta: Same as hookFile ☝️

thibmaek:
Sure, think I fixed it now.

carloscuesta:
Great! Take a look at the build @thibmaek the tests are failing 🚨

thibmaek:
Woops, you're right! Totally forgot to update the test property but should be fine now. Also added a path check for `gitmoji#remove`

Issue #56, opened 2017-07-03
Remember signed commits option globally


sivakumar-kailasam:
Hello @carloscuesta!

Signing commits is not a widespread practice yet, so it would help to remember it as a global config. If its set, use that option instead of prompting on each run.

carloscuesta:
Hey @sivakumar-kailasam 😎 

I think it's a good idea, will take a look!

PR #69, opened 2017-10-12
Add signed config

kykungz:
## Description
- Added default signed commit option in <gitmoji -g>
- Use constants in GitmojiCli `config()`

Issue: #56

## Tests
- [x] All tests passed.

### DIFF
--- a/src/gitmoji.js
+++ b/src/gitmoji.js
@@ -17,9 +17,11 @@ class GitmojiCli {
   constructor (gitmojiApiClient, gitmojis) {
     this._gitmojiApiClient = gitmojiApiClient
     this._gitmojis = gitmojis
+    /* set defaults */
     if (config.getAutoAdd() === undefined) config.setAutoAdd(true)
     if (!config.getIssueFormat()) config.setIssueFormat(constants.GITHUB)
     if (!config.getEmojiFormat()) config.setEmojiFormat(constants.CODE)
+    if (!config.getSignedCommit()) config.setSignedCommit(constants.NONE)
   }
 
   config () {
@@ -27,6 +29,7 @@ class GitmojiCli {
       config.setAutoAdd(answers.autoAdd)
       config.setIssueFormat(answers.issueFormat)
       config.setEmojiFormat(answers.emojiFormat)
+      config.setSignedCommit(answers.signedCommit)
     })
   }
 
@@ -88,7 +91,15 @@ class GitmojiCli {
     return this._fetchEmojis()
       .then((gitmojis) => prompts.gitmoji(gitmojis))
       .then((questions) => {
+
+        if (config.getSignedCommit() !== constants.NONE) {
+          questions = questions.filter(q => q.name !== 'signed')
+        }
+
         inquirer.prompt(questions).then((answers) => {
+          if (config.getSignedCommit() !== constants.NONE) {
+            answers.signed = config.getSignedCommit() === 'yes'
+          }
           if (mode === constants.HOOK_MODE) this._hook(answers)
           return this._commit(answers)
         })
--- a/src/gitmoji.js
+++ b/src/gitmoji.js
@@ -153,18 +153,6 @@ class GitmojiCli {
     })
   }
 
-  _isCommitSigned (sign) {
-    let signed
-
-    if (sign) {
-      signed = '-S'
-    } else {
-      signed = ''
-    }
-
-    return signed
-  }
-
   _isAGitRepo () {
     return parentDirs(process.cwd())
       .some((directory) => pathExists.sync(path.resolve(directory, '.git')))
--- a/src/gitmoji.js
+++ b/src/gitmoji.js
@@ -17,11 +17,10 @@ class GitmojiCli {
   constructor (gitmojiApiClient, gitmojis) {
     this._gitmojiApiClient = gitmojiApiClient
     this._gitmojis = gitmojis
-    /* set defaults */
     if (config.getAutoAdd() === undefined) config.setAutoAdd(true)
     if (!config.getIssueFormat()) config.setIssueFormat(constants.GITHUB)
     if (!config.getEmojiFormat()) config.setEmojiFormat(constants.CODE)
-    if (!config.getSignedCommit()) config.setSignedCommit(constants.NONE)
+    if (config.getSignedCommit() === undefined) config.setSignedCommit(true)
   }
 
   config () {
@@ -91,13 +90,7 @@ class GitmojiCli {
     return this._fetchEmojis()
       .then((gitmojis) => prompts.gitmoji(gitmojis))
       .then((questions) => {
-        if (config.getSignedCommit() !== constants.NONE) {
-          questions = questions.filter(q => q.name !== 'signed')
-        }
         inquirer.prompt(questions).then((answers) => {
-          if (config.getSignedCommit() !== constants.NONE) {
-            answers.signed = config.getSignedCommit() === 'yes'
-          }
           if (mode === constants.HOOK_MODE) this._hook(answers)
           return this._commit(answers)
         })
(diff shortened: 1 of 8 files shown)


carloscuesta:
Hey @kykungz thanks for sending a pull request!

src/gitmoji.js
```diff
@@ -88,7 +91,13 @@ class GitmojiCli {
     return this._fetchEmojis()
       .then((gitmojis) => prompts.gitmoji(gitmojis))
       .then((questions) => {
+        if (config.getSignedCommit() !== constants.NONE) {
```
carloscuesta: You don't need to make this, the default value has to be set on the constructor, then you can remove completely the NONE constant.

src/gitmoji.js
```diff
@@ -88,7 +91,13 @@ class GitmojiCli {
...
+          questions = questions.filter(q => q.name !== 'signed')
+        }
         inquirer.prompt(questions).then((answers) => {
+          if (config.getSignedCommit() !== constants.NONE) {
```
carloscuesta: Same as the other block, you won't need it if you set the default value on the constructor 👍

src/prompts.js
```diff
@@ -21,6 +21,12 @@ const config = [
     choices: [
       { name: ':smile:', value: 'code' }, { name: '😄', value: 'emoji' }
     ]
+  },
+  {
+    name: constants.SIGNED_COMMIT,
+    message: 'Choose default answer for Signed commit',
+    type: 'list',
```
carloscuesta: Why you don't use type `confirm`? You'll get a boolean (true/false) that it's easier to use and makes more sense! If you think it, you can sign or not sign a commit so that's a boolean

src/prompts.js
```diff
@@ -21,6 +21,12 @@ const config = [
     choices: [
       { name: ':smile:', value: 'code' }, { name: '😄', value: 'emoji' }
     ]
+  },
+  {
+    name: constants.SIGNED_COMMIT,
+    message: 'Choose default answer for Signed commit',
```
carloscuesta: `Enable signed commits`

test/gitmojiCli.spec.js
```diff
@@ -60,6 +60,11 @@ describe('config module', () => {
     config.setIssueFormat('github')
     expect(config.getIssueFormat()).toMatchSnapshot()
   })
+
+  it('should match for setSignedCommit and getSignedCommit', () => {
+    config.setSignedCommit('no')
```
carloscuesta: `true` & `false`

kykungz:
Thanks for the comment! From my understanding, you want the signed commit option in the configuration menu to have 2 values `true` & `false` ? Can you explain more, because what I thought was this option would ask to set the default value of the `Signed commit: (Y/n)` option. The `none` option means do not set any default value and always ask for `Signed commit: (Y/n)`, but `yes` and `no` will prevent asking for Signed commit and automatically answer with the chosen value.

carloscuesta:
When you specify a type `list` on the config prompts you're giving the user three options to choice: 

screen shot 2017-10-12 at 22 23 21

If you set the type to confirm, it will only accept (y/N) values that basically under the hood are `true` or  `false`. Such as the `Enable git add` question. Then you can remove the `no`, `yes`, `none` values. 

After that on the constructor, if there's no saved value for the `signed` question (undefined) you set it to true.

kykungz:
Oh I got it! I guess I understood it incorrectly, I thought that if you set this option to `yes` or `no`, when you use `gitmoji -c` it won't show the `Signed commit: (Y/n)` option at all, so that's why I put `none` option with it. So it should be an option to set the default value so that when you see `Signed commit: (Y/n)` and you press Enter without inputing any thing, it will use the default value that has been set, right? Btw, aren't the default value suppose to be 'no' ?

carloscuesta:
No, that depends on the type of the prompt, you set it to list so the values are the ones you've specified. Which in this case is an array of strings. Setting the question to confirm will generate a (y/n) question automatically without providing any value.

If you follow the comments I think you'll be able to fix it 👍

kykungz:
I understand the setting `type: confirm` but I'm a bit confused of what this option is for. In the comment, you said to set the message to `Enable signed commits`, so is this option is to enable/disable prompting for `Signed commit: (Y/n)` ? Or is it should be an option for setting default value (yes or no).
If this is an option to enable/disable prompting for `Signed commit: (Y/n)`, when it is set to `false` what should be the default value (signed or not signed) ?
Sorry for bothering, I'm still new to this PR thingy.

kykungz:
https://github.com/carloscuesta/gitmoji-cli/issues/56
I think what this user wanted is a way to set an answer to `Signed commit: (Y/n)` so that it wouldn't ask at all. So the type of configuration should be `confirm` I agree, and `Signed commit: (Y/n)` should never show again. But what if I want it to comeback? As far as I know, gitmoji doesn't have option to reset the configuration?

carloscuesta:
Yes, you have the option to configure gitmoji again by running `gitmoji -g`

kykungz:
I mean what if I want it to prompt `Signed commit: (Y/n)` every time again, what should I do. Because the config is `yes` or `no`, which both will prevent `Signed commit: (Y/n)` for showing when `gitmoji -c`

carloscuesta:
It's a global option, hence it should not be prompted every time you run `gitmoji -c`. You can set up your preference any time you want with `gitmoji -g` that is mean to be used to set up **global** preferences.

kykungz:
Okay so should I remove the `Signed commit: (Y/n)` from prompt.js?

carloscuesta:
No, you have to change the type of the question, take a look at the review

kykungz:
Something like this?

kykungz:
Should I send a new pull request? the commits for this one is a bit messy.

carloscuesta:
Don't worry we can continue on this, at the end all the commits are going to be squashed

kykungz:
Please review changes

test/gitmojiCli.spec.js
```diff
@@ -81,28 +81,26 @@ describe('gitmoji module', () => {
...
       expect(gitmojiCli._commit(stubs.prompts)).toMatchSnapshot()
     })
 
     it('should match for the commit snapshot with the given prompts', () => {
       config.setIssueFormat('jira')
+      config.setSignedCommit(true)
       expect(gitmojiCli._commit(stubs.promptsJira)).toMatchSnapshot()
     })
+
+    it('should match for the commit snapshot with the given prompts', () => {
```
carloscuesta: We don't need to create this specific test, just `setSignedCommit` to false on the jira one 👍

test/gitmojiCli.spec.js
```diff
@@ -81,28 +81,26 @@ describe('gitmoji module', () => {
   describe('commit', () => {
     it('should match for the commit snapshot with the given prompts', () => {
       config.setIssueFormat('github')
+      config.setSignedCommit(true)
       expect(gitmojiCli._commit(stubs.prompts)).toMatchSnapshot()
     })
 
     it('should match for the commit snapshot with the given prompts', () => {
       config.setIssueFormat('jira')
+      config.setSignedCommit(true)
```
carloscuesta: change it to `false` and update the snapshots!

test/stubs.js
```diff
@@ -23,6 +23,14 @@ const promptsJira = {
   signed: true
 }
 
+const promptsUnsigned = {
```
carloscuesta: Those  stubs should be removed when deleting the test ✅

test/stubs.js
```diff
@@ -31,5 +39,6 @@ module.exports = {
   gitmojiApiClient,
   prompts,
   promptsJira,
+  promptsUnsigned,
```
carloscuesta: Also this export ☝️

kykungz:
Thank you so much for guiding me through my first PR on a brilliant project!

carloscuesta:
No problem, glad you enjoyed it, and thanks for contributing to gitmoji-cli @kykungz 

Going to make a release now :tada:

Issue #86, opened 2018-01-23
Gitmoji init command is not recursive inside a git folder.


waghcwb:
Hello @carloscuesta!

Gitmoji init command is not recursive inside a git folder.

Issue: 
If we have a git repo, then enter some folder inside that repo and run `gitmoji --init`, `gitmoji-cli` tell us that the hook was sucessfully created
`gitmoji commit hook created successfully.`.

But it doesn't created the hook. 

**OS**: Deepin OS 15.5 (Linux, Debian based)
**gitmoji-cli -v**: 1.8.7
**Node version**: 9.2.0

**Error Message**:  
```
ERROR: Error: ENOENT: no such file or directory, open '~/.server/gitmoji-test/test/.git/hooks/prepare-commit-msg'
```

![gitmoji](https://i.imgur.com/Z9RFlyW.png)

PR #90, opened 2018-03-16
Create hook from anywhere in git repo

CGA1123:
## Description

Use `git rev-parse --absolute-git-dir` to find `.git/` folder

Issue: #86

## Tests

- [x] All tests passed.


src/gitmoji.js
```diff
@@ -5,6 +5,7 @@ const inquirer = require('inquirer')
 const parentDirs = require('parent-dirs')
 const path = require('path')
 const pathExists = require('path-exists')
+const { exec } = require('child_process')
```
carloscuesta: Could you use `execa` instead of child_process?

Issue #91, opened 2018-04-10
Can not use gitmoji behind an HTTP proxy


teid:
Hello @carloscuesta!

I ran into an issue when using gitmoji-cli through a HTTP corporate proxy.
It seems to be related to this issue on Axios which has been stuck couple months: https://github.com/axios/axios/pull/959

Would you accept a PR to migrate the `gitmojiApiClient` to another http client ?

Issue: 
When using gitmoji-cli through a HTTP proxy, Axios tries to connect to the proxy as an HTTPS proxy. 

You can reproduce the issue by setting up a squid proxy:
```
docker run --name squid -d --restart=always --publish 3128:3128 sameersbn/squid:3.3.8-23
export HTTPS_PROXY=http://localhost:3128
gitmoji -c
```

**OS**: Linux
**gitmoji-cli -v**: 1.8.8
**Node version**: v9.11.1

**Error Message**: 
> ERROR: Network connection not found - EPROTO

aneveux:
Hello @carloscuesta :)

I hope you're doing fine!

I'm a huge fan of gitmoji and I sadly ran in the same problem as @teid while using gitmoji-cli behind a corporate proxy. I luckily found this issue but I'm so sure about the status of it, do you have any information about that topic?

Thanks a lot, and have an awesome day!

cchang62:
Hi, @carloscuesta  Your gitmoji is such a great work but I also suffer the proxy issue. I think proxy settings are correct. Thank you.

xurizaemon:
It does look like axios/axios#659 has been merged and released in their 0.19.0, from all the snyk mentions in that issue. May be that we now can just update to that release?

PR #92, opened 2018-04-14
Use the HTTPS_PROXY env var

teid:
## Description

Handle the HTTPS request to Github over an HTTP proxy if the HTTPS_PROXY var is set

Issue: #91 

## Tests

- [x] All tests passed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.log
 node_modules/
 coverage/
+.vscode
--- a/cli.js
+++ b/cli.js
@@ -36,7 +36,7 @@ const cli = meow(`
   }
 })
 
-const gitmojiCli = new GitmojiCli(utils.gitmojiApiClient)
+const gitmojiCli = new GitmojiCli()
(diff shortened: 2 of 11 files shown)


teid:
Hi @carloscuesta!, this PR makes the cli to work over an HTTP proxy :)
It uses the `HTTPS_PROXY` and `https_proxy` environment variables to enable the feature (with the agent: https://github.com/kevva/caw)

However I did not not implements the connection over an HTTPS proxy.
I think it should not be too hard, but it requires couple more hours. And I do not have the need to do it right now, maybe we can delay this feature until the usecase comes up ?

carloscuesta:
Hi @teid 

A lot has gone through since the opening of this PR, It just got stalled and a lot of refactors happened through time 😞. I'm so sorry but I don't feel like I can make the needed work to merge this right now, so I'm closing it

I would really want to thanks you for the time and effort you put in this PR, I really appreciate it 🙏

PR #99, opened 2018-06-06
:sparkles: Add sign-off support

rikakomoe:
Introduce a config to add the Signed-off-by field (-s) while committing

Signed-off-by: LI JIAHAO <lijiahao99131@gmail.com>

## Description

Adding the "Signed-off-by" field in commits might be useful in some project practice. Not enabled by default since most projects don't use it.

Issue: #

## Tests

- [x] All tests are passing within the default configuration.


carloscuesta:
Hello @rikakomoe 

Thanks for sending a PR. Please consider opening an issue before creating any changes!

I think we don't need to support for "Sign-off" since you can sign your commits with GPG. I know it's not the same, but for me it's much better to sign with GPG, rather than having a "Sign-off" text on the commit message.

Issue #64, opened 2017-08-23
Unable to create .git/index.lock


svenlombaert:
Hello @carloscuesta 😎!

Since a while now, every time I commit (I have the pre-commit hook installed), after it, I get this message:

```
ERROR: fatal: Unable to create '/Projects/pathname/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.
```
I've tried removing and installing the pre-commit hook, removed gitmoji-cli completely and updated to the most recent version - nothing helps. I encounter the same problem without pre-commit hook and just using the gitmojicommand. When I just commit normally (without gimoji) everything works fine 👌 .

Any ideas?

KR

Seaony:
I also met the same problem, did you find the solution?

carloscuesta:
Related with #9 #10 #62

svenlombaert:
MacOS Sierra, Version 10.12.6 (16G29)

Thanks!

samirdhebar:
I also am having the same issue on Linux regarding this commit issue. I reviewed the earlier fix but sadly to no avail :cry:

LBreda:
Same here, on Linux. The lockfile doesn't actually exist.

gluons:
Me too. On Kubuntu 17.04. It happened sometime.

![_002](https://user-images.githubusercontent.com/4688092/30250443-9556e28e-9678-11e7-96aa-0ff90301f577.png)

gluons:
Did anyone found the cause of this problem?  
I frequently face this problem but I can't find why it happened. 🤔  
Try to find `index.lock` file in my `.git` folder but it doesn't exist.  😵

carloscuesta:
As the error messages says, this problem comes from: 

> Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.

I've been investigating and I haven't found a solution yet. It's really weird because it's hard to reproduce the error. IMHO I think the issue is related with use of the `hook` option

https://github.com/carloscuesta/gitmoji-cli/blob/ea99b5166c2d642d72be7f536d69127cecb99409/src/gitmoji.js#L108

[...]

PR #109, opened 2018-07-09
Fix thow 'Unable to create .git/index.lock'

muwoo:
## Description

`git commit` should used after `git add .`

Issue: #64 

## Tests

- [x] All tests passed.


muwoo:
Help Wanted.

I have no idea about writing the test.

src/gitmoji.js
```diff
@@ -149,13 +149,17 @@ class GitmojiCli {
 
     if (config.getAutoAdd()) {
       execa.stdout('git', ['add', '.'])
-        .then((res) => console.log(chalk.blue(res)))
-        .catch((err) => this._errorMessage(err.stderr))
+        .then((res) => {
+          console.log(chalk.blue(res))
```
carloscuesta: ⚠️ Remove that console.log

src/gitmoji.js
```diff
@@ -149,13 +149,17 @@ class GitmojiCli {
...
+          return execa.shell(commit)
```
carloscuesta: You can return this directly from the arrow function

muwoo: Thanks a lot! I've changed it.

PR #112, opened 2018-07-25
:children_crossing: Ignore hook when rebasing

Winangel:
The hook breaks the git rebase process. Modified script to not apply the hook when we are in rebase mode. See https://gmurphey.github.io/2013/02/02/ignoring-git-hooks-when-rebasing.html#.W1iRxtgzYWo for details

## Description

When rebasing with the gitmoji hook set, the hook breaks the process. In order to make it compatible with rebasing, I use the trick of gmurphey (see:  https://gmurphey.github.io/2013/02/02/ignoring-git-hooks-when-rebasing.html#.W1iRxtgzYWo) so that we can do rebase as usual.

## Tests

Tests are broken due to snapshot comparison with the previous Constant.js file. If the Pull request get approved I will modify the tests so that they pass.

- [ ] All tests passed.


chdeliens:
@carloscuesta could there be a regression in recent versions? Got a fresh Manjaro install here and gitmoji is triggering on rebase! Did not happen before :/ any clue? How can I help?

ml-milan-vit:
Same here, this seems to be a regression. I now have to disable Gitmoji before doing a rebase.

david-mears-dfe:
Same as ml-milan-vit

carloscuesta:
This sould have been caused by another version bump I guess 😕

carloscuesta:
I removed the `hook` changes because everytime we made a modification we broke something else. We have a PR #366 to fix that but we need to improve the way we test this in order to prevent breaking the hook every time we change it

Issue #104, opened 2018-06-23
Error when using -i option


GitSquared:
Hello @carloscuesta! (nice issue template by the way, i might just steal it)

I can't seem to setup the gitmoji hook using `gitmoji-cli`.

Issue: 
- create a new git repo
- `cd` into the thing
- run `gitmoji -i`
- get error message below
- cry

**OS**: Debian 9
**gitmoji-cli -v**: 1.9.0
**Node version**: v9.11.1

**Error Message**: `ERROR: Error: ENOENT: no such file or directory, open '--absolute-git-dir/hooks/prepare-commit-msg'`

carloscuesta:
Hey @GitSquared thanks for the report that might be related with #90 @CGA1123

GitSquared:
@Keyzou thanks for the help. I'll try this when i get home.

PR #129, opened 2018-09-16
Support older git versions (< 2.13 - for debian stable/stretch)

nathan818fr:
## Description

The `--absolute-git-dir` option has been added to git 2.13.0. This option is used for gitmoji hook install and remove.
The stable version of debian (stretch) uses git 2.11, which prevents some members of my team from using gitmoji (they can't use stretch-backports currently).

This fix detect the current git version and use the best available method/
- Git >= 2.13: Use `--absolute-git-dir`
- Git < 2.13: Use `--git-dir` then `path.resolve` to canonicalized the absolute path

Issue: #104

## Tests

- [x] All tests passed.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -11,6 +11,7 @@ const ISSUE_FORMAT = 'issueFormat'
 const JIRA = 'jira'
 const SIGNED_COMMIT = 'signedCommit'
 const TITLE_MAX_LENGTH_COUNT = 48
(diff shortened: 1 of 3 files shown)


carloscuesta:
Hey @nathan818fr 

Before creating any PR please create an issue first.

I don't think it's a good mindset to try supporting old versions of git. I think we should embrace users to try to stay up to date with the software updates that are coming. 

Thanks for sending a PR!

PR #192, opened 2019-07-19
✨ Add extraArgs option

zaber-paul:
## Description

This can be used to provide extra arguments to the 'git commit' command, e.g. '--fixup HEAD'

## Tests

Snapshot was updated with `yarn jest --updateSnapshot`
- [x] All tests passed.

### DIFF
--- a/src/commands/commit/prompts.js
+++ b/src/commands/commit/prompts.js
@@ -19,7 +19,8 @@ export type Answers = {
   gitmoji: string,
   scope?: string,
   title: string,
-  message: string
+  message: string,
+  extraArgs: string
 }
 
 export default (gitmojis: Array<Gitmoji>): Array<Object> => [
@@ -62,5 +63,9 @@ export default (gitmojis: Array<Gitmoji>): Array<Object> => [
     name: 'message',
     message: 'Enter the commit message:',
     validate: guard.message
+  },
+  {
+    name: 'extraArgs',
+    message: 'Extra commit args (e.g. --fixup HEAD):'
   }
 ]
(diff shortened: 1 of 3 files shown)


src/prompts.js
```diff
@@ -67,6 +67,11 @@ const gitmoji = (gitmojis) => {
       name: 'reference',
       message: 'Issue / PR reference:',
       validate: (value) => guard.reference(value, configVault.getIssueFormat())
+    },
+    {
+      name: 'extraArgs',
+      message: 'Extra commit args (e.g. --fixup HEAD):',
+      validate: guard.extraArgs
```
carloscuesta: The `guard.extraArgs` method does not exists in the `guard.js` file, maybe you forgot to add it?

test/__snapshots__/gitmojiCli.spec.js.snap
```diff
@@ -132,6 +132,11 @@ Array [
     "name": "reference",
     "validate": [Function],
   },
+  Object {
+    "message": "Extra commit args (e.g. --fixup HEAD):",
+    "name": "extraArgs",
+    "validate": undefined,
```
carloscuesta: Referring to the comment I posted, that's why this is undefined in the snapshot

zaber-paul:
I removed the validation function since any options (and future git commit options) should be allowed. If you enter incorrect text for `extraArgs`, the commit will fail with a message from `git`. Also, updated the snapshots.

I only ever use the `--fixup` option here (personally), so I'd be OK to rewrite this PR as a dedicated prompt for fixup reference if you prefer (it could have a validation function matching SHAs and branch names). Or, if you prefer to merge I think this PR is in a mergeable state now.

carloscuesta:
Hey @zaber-paul 

Thanks for the work you put into this PR, I really appreciate that❤️ 

I want to focus on making the cli simpler and not introduce a lot of new features.

Thanks!

PR #221, opened 2019-10-25
Set current values as default in `gitmoji -g`

nathan818fr:
Set the current config values as default when using `gitmoji -g`.
This is very useful when you want to change a single value.

### DIFF
--- a/src/commands/config/prompts.js
+++ b/src/commands/config/prompts.js
@@ -1,4 +1,6 @@
 // @flow
+import configurationVault from '../../utils/configurationVault'
+
 export const CONFIGURATION_PROMPT_NAMES = {
   AUTO_ADD: 'autoAdd',
   EMOJI_FORMAT: 'emojiFormat',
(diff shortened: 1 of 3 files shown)


carloscuesta:
Hey @nathan818fr this could be useful when you only want to change a single value, but what happens when you want to change more than one ? You should be removing the current values from the prompts ?

nathan818fr:
No. For boolean defaul value is only used when imput is empty, and for select the value is just pre-selected.
Peek 2019-11-06 08-21

carloscuesta:
Oh! Ok I understand now could you make a rebase ? 🙏

nathan818fr:
Yep, It's done!

src/commands/config/prompts.js
```diff
@@ -11,11 +13,12 @@ export const EMOJI_COMMIT_FORMATS = {
   EMOJI: 'emoji'
 }
 
-export default [
+export default () => [
```
carloscuesta: Why do we need to convert this into a function?

nathan818fr: This object is not longer immutable since it depend on `configurationVault` values. Changing it to a function ensures that it always use the current config.

eg:
```
configurationVault.setAutoAdd(true)
// configurationPrompts()[0].default is true
configurationVault.setAutoAdd(false)
// configurationPrompts()[0].default is false
```

PR #222, opened 2019-10-27
♻️ Cli refactor

carloscuesta:
This PRs is a major refactor for `gitmoji-cli`. I'm changing the architecture of the code and improving aswell the UX and the Developer UX for contributors.

This will be released in a major version since we're dropping support for Node.js `6` and `8`. The minimum required version will be `10`.


carloscuesta:
Whole code is refactored, we need to made all the tests again 👀

PR #247, opened 2019-12-04
Ignore hook if not creating a new commit, from a TTY

hlerebours:
## Description
Ignore hook if not creating a _new_ commit, from a TTY

It solves and supersedes https://github.com/carloscuesta/gitmoji-cli/pull/112: Ignore hook when rebasing.

## Tests
- [x] All tests passed.

### DIFF
--- a/src/commands/hook/hook.js
+++ b/src/commands/hook/hook.js
@@ -4,8 +4,11 @@ const HOOK: Object = {
   PATH: '/hooks/prepare-commit-msg',
   CONTENTS: `#!/bin/sh
 # gitmoji as a commit hook
-exec < /dev/tty
-gitmoji --hook $1
+if test -t 1 && ! grep -q -m 1 '^[^#]' $1; then
+  # it has been invoked from a tty and no commit message is already set
+  exec < /dev/tty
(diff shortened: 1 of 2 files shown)


src/commands/hook/hook.js
```diff
@@ -2,9 +2,14 @@
 const HOOK: Object = {
   PERMISSIONS: 0o775,
   PATH: '/hooks/prepare-commit-msg',
-  CONTENTS:
-    '#!/bin/sh\n# gitmoji as a commit hook\n' +
-    'exec < /dev/tty\ngitmoji --hook $1\n'
+  CONTENTS: `#!/bin/sh
+# gitmoji as a commit hook
+if test -t 1 && ! grep -q -m 1 '^[^#]' $1; then
```
carloscuesta: Could you explain a little bit more this line ?

Seems a little bit tricky and hacky to my understand 😂

hlerebours: I rephrased the comment that was below the condition.
The first condition means we are in a TTY (the hook has not been triggered from a program but likely from a user action), the second one means there is no comment already set.

PR #256, opened 2019-12-15
Add scope template

vhoyer:
## Description

> **TL;DR:**
> I wanted to change the way `gitmoji-cli` added scope to commit, so I added an option to do it

I started using gitmoji without cli, then, I saw about the convetion some people use about adding a scope to the commit instead of mention it in the actuall commit message. I started using the model [`:sparkles: [nvim] open terminal at vertical split`](https://github.com/vhoyer/dotfiles/commits). After some time I started using the CLI, because it is more convinient. Now, 3 hours ago, I discovered that there was the option to config `gitmoji-cli` to prompt for the scope part of the commit. I saw the model in current use (`(scope): `) and didn't like it since I was using this other template (`[scope] `).

Motivated by this I added the option in `--config` to set a custom template for the scope part of the commit in the interactive commit insertion provided by this repo :D

I'd very much appreciate to start using this feature soon on my repo with the official package :D

Issue: `#what issue?`

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/commit/withClient/index.js
+++ b/src/commands/commit/withClient/index.js
@@ -9,7 +9,11 @@ import { type Answers } from '../prompts'
 
 const withClient = async (answers: Answers) => {
   try {
-    const scope = answers.scope ? `(${answers.scope}): ` : ''
+    const scopeTemplate = configurationVault.getScopeTemplate()
+
+    const scope = answers.scope
+      ? scopeTemplate.replace('%s', answers.scope)
+      : ''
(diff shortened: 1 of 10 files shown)


carloscuesta:
Hey @vhoyer 

I don't think that is a big deal the usage of `() vs []` I prefer to keep the `()` since this is more common on [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope) I don't think we need to add more options to custom the scope character, I think adds a lot of code complexity just to change a character

Thanks!

Issue #147, opened 2018-11-27
[Feature Request] Add word tags/fuzzy search to improve searching


m-roberts:
Hello @carloscuesta!

I was wondering if it would be possible to consider adding synonyms to words included in the full text descriptions to improve results. Allowing for a partial match would be very helpful too. Being able to to search for synonyms of the emoji string would also be helpful.

For example:

| Search For | Result Shown             |
|------------|--------------------------|
| comment    | Documenting source code. |
| Jenkins    | Adding CI build system.   |
| add asset    | Adding or updating assets.   |
| move    | Moving or renaming files.   |
| emergency    | Critical hotfix.   |
| poo    |  Writing bad code that needs to be improved.   |

In particular, because it is [recommended to use imperative mood in the subject line of a Git commit](https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53#rules-for-a-great-git-commit-message-style) (i.e. A properly formed Git commit subject line should always be able to complete the following sentence: `If applied, this commit will <your subject line here>`), allowing for searching of words such as `move` instead of `moving` is helpful.

This would make finding the most appropriate emoji a lot easier.

carloscuesta:
Hey @m-roberts 

I agree with it, we can improve a lot the search of emojis with the cli!

PR #267, opened 2019-12-23
Add fuzzy search

adriencaccia:
## Description

Use [fuzzy](https://www.npmjs.com/package/fuzzy) search.
This will for example make such a search possible:
image

Issue: #147 

## Tests

- [x] All tests passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "chalk": "^3.0.0",
     "conf": "^6.1.0",
     "execa": "^4.0.0",
+    "fuse.js": "^3.4.6",
     "inquirer": "^7.0.0",
     "inquirer-autocomplete-prompt": "^1.0.1",
     "meow": "^6.0.0",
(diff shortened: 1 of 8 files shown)


adriencaccia:
hey @carloscuesta, I moved the `fuzzy` filtering to a separate utils file and tested it, but the coverage diff is still not high enough :cry:.
Do you have an idea on how to get it higher ?
Thanks !

package.json
```diff
@@ -40,6 +40,7 @@
     "chalk": "^3.0.0",
     "conf": "^6.1.0",
     "execa": "^4.0.0",
+    "fuzzy": "^0.1.3",
```
carloscuesta: Can we replace this dependency for something else that is more maintained ?

https://www.npmjs.com/package/fuse.js

src/commands/search/index.js
```diff
@@ -1,16 +1,14 @@
 // @flow
+import fuzzy from 'fuzzy'
```
carloscuesta: This import is unused 🤔

adriencaccia:
Thanks for the review !
Sorry I didn't follow the contributing guidelines before submitting the PR, I thought that since #147 existed you were waiting for someone to create it :wink:.
I used fuse.js instead of fuzzy and cleaned the code a bit !

PR #268, opened 2019-12-24
Improvements for hook mode

adriencaccia:
## Description

Two improvements:

- Prevent the hook from being run when doing an interactive rebase
- Allow users to pre-fill the commit title and message when running `git commit -m "My custom title" -m "My custom message"`
image
image
Result when pressing `Enter` on title and message steps:
image

This last feature will allow users to keep the commits title and message in their shell history, which I found greatly improves the cli's experience.

## Tests

- [x] All tests passed.


carloscuesta:
Fixed in #254

Issue #291, opened 2020-01-23
[Feature Request] Add --verbose option


BenjaminWolfe:
Hi @carloscuesta!

I love `gitmoji` and `gitmoji-cli`. Thank you for making it happen.

I have my own pre-commit hook that makes changes to files and messages the user in the process, so that there's a record and some awareness of what's changed.

I'll admit the _way_ I do it is a bit hacky; I know R far better than I know bash, so I have the pre-commit hook call R to run a script. But it gets the job done, and it also messages the user about it.

That is to say, when I use a simple `git commit` it messages the user. But when I use `gitmoji-cli -c`, there is no messaging.

I think maybe the messaging ends up in `stderr`. For example, I tested what would happen if I had my pre-commit hook always exit with non-zero status. I learned via that test that at least the hook was working... but I also saw my R messaging come through in `stderr`. But when I changed it to exit successfully, alas, I lost my messaging.

Is there a way to get that back? As I suggested in the title, maybe a `--verbose` or `-v` option?

Jeremie-Chauvel:
I second this, gitmoji-cli eats hooks outputs which can complicate understanding what's happening (cli looks blocked).
Instead of a verbose option, I would recommended a stricter handle of stderr/stdout to avoid blocking other hooks outputs.

carloscuesta:
Definitely this would be a great improvement into the cli.

Would we need to show the output from pre-commit hooks always or only whenever they fail ?

Jeremie-Chauvel:
I would say always, I can take a look when I have some time

carloscuesta:
Sure @Jeremie-Chauvel we can discuss the implementation of the feature on this issue!

Jeremie-Chauvel:
Hi @carloscuesta,
will look into this this week if you are still interested ?

carloscuesta:
> Hi @carloscuesta,
> 
> will look into this this week if you are still interested ? 

Hey @Jeremie-Chauvel!

For sure feel free to work on this issue 😍

That would be really helpful, do you have an idea about the implementation you want to follow?

Jeremie-Chauvel:
Not sure, I will look into it and ping you to challenge the implementation if it's alright ?

carloscuesta:
Sounds good! No worries 🙌🏼

Jeremie-Chauvel:
From a quick look into it, execa support to disable the output buffer, to stream the process, tested in a quick POC,
[...]

PR #344, opened 2020-04-01
:sparkles: show output from git commit

codedrift:
## Description
Currently the the output from pre-commit hooks is not printed to stdout. 
That makes it hard to spot errors as they appear and also leaves you hanging while the commit hook is executed.
The commit message is also lost when the pre-commit hook fails. This is a bit annoying because you have to re-type it afterwards.

This change prints the output and the whole command in the case of an error.

Issue: #291 is somewhat related

## Tests
Some tests were failing before this change and a lot more after this change. I did not bother fixing them since it would depend on the actual implementation of this feature.

### DIFF
--- a/src/commands/commit/withClient/index.js
+++ b/src/commands/commit/withClient/index.js
@@ -26,18 +26,27 @@ const withClient = async (answers: Answers) => {
 
     if (configurationVault.getAutoAdd()) await execa('git', ['add', '.'])
 
-    const { stdout } = await execa('git', [
+    const extraMessageArgs = answers.message
+      ? ['-m', `'${answers.message}'`]
+      : []
+
+    const args = [
       'commit',
       ...isSigned,
       '-m',
-      title,
-      '-m',
-      answers.message
-    ])
+      `'${title}'`,
+      ...extraMessageArgs
+    ]
 
-    console.log(stdout)
+    await execa('git', args, { stdio: 'inherit', shell: true })
   } catch (error) {
-    console.error(error)
+    const { exitCode, command } = error
+    console.log('')
+    console.log(
+      chalk.bold('Commit failed with code'),
+      chalk.bold(chalk.red(exitCode))
+    )
+    console.log(chalk.bold('Command:'), chalk.cyan(command))
   }
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,8449 @@
+{
+  "name": "gitmoji-cli",
+  "version": "3.2.3",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/cli": {
+      "version": "7.8.4",
+      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.8.4.tgz",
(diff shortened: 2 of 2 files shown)


Jeremie-Chauvel:
@codedrift maybe you could avoid committing the `package-lock.json` as the project use yarn.

src/commands/commit/withClient/index.js
```diff
@@ -26,18 +26,27 @@ const withClient = async (answers: Answers) => {
...
-    ])
+      `'${title}'`,
+      ...extraMessageArgs
+    ]
 
-    console.log(stdout)
+    await execa('git', args, { stdio: 'inherit', shell: true })
   } catch (error) {
-    console.error(error)
+    const { exitCode, command } = error
```
Jeremie-Chauvel, line 43: why not destruct in catch ? 

```js
} catch ({ exitCode, command }) {
```

src/commands/commit/withClient/index.js
```diff
@@ -26,18 +26,27 @@ const withClient = async (answers: Answers) => {
...
+    console.log('')
+    console.log(
+      chalk.bold('Commit failed with code'),
+      chalk.bold(chalk.red(exitCode))
+    )
+    console.log(chalk.bold('Command:'), chalk.cyan(command))
```
Jeremie-Chauvel, line 49: it's important to keep `console.error` instead of `console.log` since `console.error` redirect to `stderr` vs `stdout`

```js
console.error(chalk.bold(`\nCommit failed with code ${chalk.red(exitCode)}`))
console.error(chalk.bold('Command:'), chalk.cyan(command))
```

src/commands/commit/withClient/index.js
```diff
@@ -26,18 +26,27 @@ const withClient = async (answers: Answers) => {
...
-      title,
-      '-m',
-      answers.message
-    ])
+      `'${title}'`,
+      ...extraMessageArgs
+    ]
 
-    console.log(stdout)
+    await execa('git', args, { stdio: 'inherit', shell: true })
```
Jeremie-Chauvel, line 41: I don't understand why you didn't just add the option argument but go the extra way to not add the extra '-m' if no commit message is passed ?

codedrift:
hey! Thanks for actually reviewing this PR 😅 Please don't get hung up on my implementation though. it was just a rough "hack" to have this feature asap

src/commands/commit/withClient/index.js
codedrift: I think the intention was to print the error as is AND print a nice message. Destructing the error in catch would eat up any extra content in the error message.

src/commands/commit/withClient/index.js
codedrift: sure

src/commands/commit/withClient/index.js
codedrift: i think "shell:true" produced an error with the previous implementation

Jeremie-Chauvel:
sure, I was mostly trying to relaunch the dialog about this feature, since I'm interested in the solved issue 😄

carloscuesta:
Hey @codedrift this PR got stuck :( I really think that this would be a great addition to the cli since it solves a current issue. I'm going to close it but feel free to reopen if you want to help with this 

Thanks ❤️

Issue #266, opened 2019-12-23
ctrl-c skip gitmoji prompts stopped working in 3.0.1


eli-symbolab:
Hello @carloscuesta!

hi! Thank you for this cool and useful program. I installed the latest 3.0.1 and noticed that the previous behavior of pressing ctrl-c to skip the gitmoji prompt has stopped working. Rolling back to 3.0.0 restored the behavior.

Issue:

Perform a git merge as usual
Run `git commit`, when a gitmoji-cli hook has been installed
The gitmoji prompt appears.
Press ctrl-c
_Expected_: skipping gitmoji, I see the standard Merge message pre-filled in my git commit message editor. Saving and exiting completes the merge.
_Actual behavior_: gitmoji-cli quits, and the commit is canceled.

**OS**: macOS Mojave 10.14.6
**gitmoji-cli -v**: 3.0.1
**Node version**: 6.13.4

carloscuesta:
That should have been caused for one of the dependencies we bumped on `3.0.1`

eli-symbolab:
This behavior is reproducible in 3.1.0 as well.

tadejs:
I confirm on 3.1.0 on Ubuntu 18.04 as well. I was relying on this behavior to make `git commit --amend` work without needing to retype the commit message. Is there any "official" way to skip the dialogue on commit?

eli-symbolab:
Same behavior in 3.2.0.

Tydax:
At first I thought it was a new feature rather than a bug that I was about to ask about because I find it annoying for amending commits and rebasing 😅 . Thank you for reporting this.

Same behaviour in 3.2.1.

carloscuesta:
As I pointed out this should be the reason for the problem: https://github.com/carloscuesta/gitmoji-cli/issues/266#issuecomment-568406983 if anyone is willing to take a look I would really appreciate it! 🙏

eli-symbolab:
Thank you all!!!!!

PR #352, opened 2020-04-12
Fix hook abort with ctrl-c behavior to match v3.0.0

Jeremie-Chauvel:
## Description
meow version update introduce a bug with the crl-c behavior of gitmoji: the commit is aborded instead of proceeding without gitmoji.

I pinned the version of meow to the last version without the bug.

Issue: #266

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/commit/index.js
+++ b/src/commands/commit/index.js
@@ -3,10 +3,12 @@ import inquirer from 'inquirer'
 
 import getEmojis from '../../utils/getEmojis'
 import prompts from './prompts'
-import withHook from './withHook'
+import withHook, { registerHookInterruptionHandler } from './withHook'
 import withClient from './withClient'
 
 const commit = (mode: 'client' | 'hook') => {
+  if (mode === 'hook') registerHookInterruptionHandler()
+
   return getEmojis()
     .then((gitmojis) => prompts(gitmojis))
     .then((questions) => {
--- a/test/commands/commit.spec.js
+++ b/test/commands/commit.spec.js
@@ -156,6 +156,40 @@ describe('commit command', () => {
         expect(process.exit).toHaveBeenCalledWith(0)
       })
     })
+    describe('when receiving a signal interrupt', () => {
+      it('should call process.exit(0)', async () => {
+        const warnConsoleSpy = jest.spyOn(console, 'warn').mockImplementation()
+
+        // mock process.on and process.kill to test registerHookInterruptionHandler
+        const processEvents = {}
+        jest.spyOn(process, 'on').mockImplementation((signal, cb) => {
+          processEvents[signal] = cb
+        })
+        jest.spyOn(process, 'kill').mockImplementation((pid, signal) => {
+          processEvents[signal]()
+        })
+
+        inquirer.prompt.mockImplementation(() => {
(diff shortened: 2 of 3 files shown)


carloscuesta:
Could you point it to an issue of Meow? Has it been reported?

Jeremie-Chauvel:
Investigating on Meow side, I couldn't link this to an official issue. Taking a look at the release for [v6.0.0](https://github.com/sindresorhus/meow/releases/tag/v6.0.0), I thought this could be linked to the `hardRejection` flag, however, I couldn't seem to catch the rejection, even setting the flag to `false`, hence deciding to pin Meow version to v5.0.0.

It hasn't been reported, since this use case it typically what you don't want from a cli: ctrl-c should just interrupt the cli with a return code > 0; here we want to return a success status code for the prepare-commit-msg hook.

carloscuesta:
I could understand it but we should try to fix the issue without downgrading the dependency to `5.0.0`.

Because then we won't be able to update `meow` anymore if we want to keep compatibility of this (before the issue I never knew this was a feature).

Jeremie-Chauvel:
I found a solution aligned with your goal, with the added benefit of highlighting that the ctrl-c behavior for hook mode is a feature. I will add some tests when I have some time.

carloscuesta:
That is perfect! ❤️

src/commands/commit/index.js
```diff
@@ -7,6 +7,13 @@ import withHook from './withHook'
 import withClient from './withClient'
 
 const commit = (mode: 'client' | 'hook') => {
+  // Allow to interrupt the hook without cancelling the commit
+  process.on('SIGINT', () => {
+    if (mode === 'hook') process.exit(0)
+    console.error('gitmoji-cli was interrupted')
+    process.exit(1)
+  })
```
carloscuesta, line 15: I would move this to the `withHook` file

https://github.com/carloscuesta/gitmoji-cli/blob/master/src/commands/commit/withHook/index.js

Jeremie-Chauvel: I was going for this too, but we need to register the listener before prompting the user for the questions:
which mean here https://github.com/carloscuesta/gitmoji-cli/blob/master/src/commands/commit/index.js
or higher up

Jeremie-Chauvel:
I added the mentioned test (testing an interrupt behavior is a lot more painful than I anticipated) and extracted the logic in `withHook`

Jeremie-Chauvel:
@carloscuesta Do you need anything else to merge this PR ?

test/commands/commit.spec.js
```diff
@@ -25,6 +31,7 @@ describe('commit command', () => {
           Promise.resolve(stubs.clientCommitAnswers)
         )
         getEmojis.mockResolvedValue(stubs.gitmojis)
+        registerHookInterruptionHandlerSpy.mockReset()
```
carloscuesta, line 34: Why we need to reset the mock on every test if we're not using it here ?

Can we mock and spyOn the function where we need to, in order to avoid that duplication ?

Jeremie-Chauvel: sure

eli-symbolab:
❤️ ❤️ ❤️ 
thank you!!!

Issue #397, opened 2020-06-18
[Feature discussion] Rebase and --no-edit behaviour with hook


Larandar:
Hello @carloscuesta!

I would like to reopen issue #197 with a larger discussion because, unfortunately, the current behaviour makes it difficult to use git workflows which are semi-linear or linear and require rebasing.

**Use-case:**
 
Usually, my workflow requires to branch from master and then rebase before making my MR with both compacting my history and making sure current master changes does not conflict with my branch. And then before merging rebasing a last time so that history is semi-linear.

**Possible solutions I see:**

1. Detecting --no-edit 
2. An option to disable gitmoji temporary, or during rebases.
3. Pre-filling the CLI with the current commit message if detected, making pressing enter 3 times all that is needed

**My thoughts on each solution**

1. Seem impossible currently, but maybe is-it a use-case to improve mainstream git? Knowing if the --no-edit flag is on from the prepare-commit-msg may be a legitimate question.
2. Nice but does not make the workflow that much better.
3. Seem the easier solution, and is more elegant than CTRL+C but not a lot more.

carloscuesta:
The problem is specifically with the hook right? I use rebase workflow also and have no problems using gitmoji-cli as I would normally do

carloscuesta:
We have a PR for this but is currently blocked due to: https://github.com/carloscuesta/gitmoji-cli/pull/366#issuecomment-626140318

carloscuesta:
Solved by @Jeremie-Chauvel, I'm releasing it now!

PR #366, opened 2020-04-28
Re-add hook content changes

nathan818fr:
Re-add **fixed** hook content changes:
- https://github.com/carloscuesta/gitmoji-cli/commit/d8ec4d11502425fa1eecde290ad797f8c876e86b#diff-1e69352ce20a8767469d44f951b132a9
- https://github.com/carloscuesta/gitmoji-cli/commit/d33a26bb7003efd2c3a880cc94bf8cc4da80ff30#diff-1e69352ce20a8767469d44f951b132a9

These changes were canceled by https://github.com/carloscuesta/gitmoji-cli/commit/c668e779d4866188c7fe8a3b7be8d56309e065d0 due to https://github.com/carloscuesta/gitmoji-cli/issues/330 .
The problems were caused by the use of bash syntaxes (not compatible with POSIX sh).
Now the script is fully compliant with POSIX sh!

Issue #330, opened 2020-03-11
gitmoji-cli 3.2.2 showing weird messages in terminal when commiting


xItsDanger:
Hello @carloscuesta!

Hi, I updated to 3.2.2, and everything seems to work fine, except some weird messages appearing in my terminal when commiting

Anyways, this isn't important. I can commit changes, but I think this should be fixed

This is my terminal log:

```js
andres on acer laptop at Alexis > git commit events/client/message.js 
// These are the error messages
.git/hooks/prepare-commit-msg: 6: [[: not found
.git/hooks/prepare-commit-msg: 8: [[: not found
.git/hooks/prepare-commit-msg: 6: [[: not found
.git/hooks/prepare-commit-msg: 8: [[: not found
.git/hooks/prepare-commit-msg: 6: [[: not found
.git/hooks/prepare-commit-msg: 8: [[: not found
.git/hooks/prepare-commit-msg: 6: [[: not found
.git/hooks/prepare-commit-msg: 8: [[: not found
.git/hooks/prepare-commit-msg: 6: [[: not found
.git/hooks/prepare-commit-msg: 8: [[: not found
.git/hooks/prepare-commit-msg: 6: [[: not found
.git/hooks/prepare-commit-msg: 8: [[: not found
.git/hooks/prepare-commit-msg: 6: [[: not found
.git/hooks/prepare-commit-msg: 8: [[: not found
.git/hooks/prepare-commit-msg: 6: [[: not found
.git/hooks/prepare-commit-msg: 8: [[: not found
.git/hooks/prepare-commit-msg: 6: [[: not found
.git/hooks/prepare-commit-msg: 8: [[: not found
// Now, the usual behavior continues.
? Choose a gitmoji: 🔥  - Removing code or files.
? Enter the commit title [23/48]: Removed betacmd feature
? Enter the commit message: Removed betacmd feature in message.js. Separate beta
 bot will be created
[master 334b70b] 🔥 Removed betacmd feature
 1 file changed, 31 deletions(-)
```

**OS**: Ubuntu 19.10
**gitmoji-cli -v**: 3.2.2
**Node version**: 12.16.1

**Error Message**: None

carloscuesta:
Looks like a syntax issue on the hook 😅

carloscuesta:
I'm going to revert the changes we made into the hook, since it's a delicate thing to change

carloscuesta:
[...]

PR #415, opened 2020-07-12
✨ Cancel hook when a git message is already set

Jeremie-Chauvel:
## Description

Hi @carloscuesta ,

At first I just wanted to build end to end tests over @nathan818fr [pull request](https://github.com/carloscuesta/gitmoji-cli/pull/366). However there are numerous difficulties in doing a true E2E testing of a CLI app. The deal breaker for gitmoji being that inquirer doesn't handle piping to answer. I also tried to mock some libs like inquirer to avoid the prompt but still use a classic `git commit` flow to trigger the `prepare-commit-msg` hook without success.

My last idea was to use some remotely controllable terminal to deal with the interactivity of inquirer (similar in principle to cypress for web app), but except [kitty terminal](https://sw.kovidgoyal.net/kitty/#remote-control), I don't know another terminal that implement this feature and as kitty is still recent, I figured it wasn't a good fit since stability isn't reached yet and checked out this idea as well.

Since I couldn't figure out a way of doing such e2e tests in a reliable and cross platform way, I choose to go back to the source of the problem:
- The hook is always triggered when committing even on rebase, cherry-pick, amend, etc.
- Implementing the logic of not triggering the hook when the git message is already defined should be done in a plateform agnostic way, which as shown by https://github.com/carloscuesta/gitmoji-cli/issues/330 is no small task.

Finally, I tried a small proof of concept for a pure js implementation in gitmoji and figured out it was quite an easy and elegant solution with the added benefit of being easy to test.

## Issues

Issue: https://github.com/carloscuesta/gitmoji-cli/issues/330
Issue: https://github.com/carloscuesta/gitmoji-cli/issues/397
blocked pr: https://github.com/carloscuesta/gitmoji-cli/pull/366

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/commit/withHook/index.js
+++ b/src/commands/commit/withHook/index.js
@@ -26,4 +26,16 @@ export const registerHookInterruptionHandler = () => {
   })
 }
 
+export const cancelIfMessageIsAlreadySet = () => {
+  const commitMessageFilePath = process.argv[3]
+  if (
+    fs.existsSync(commitMessageFilePath) &&
+    fs.lstatSync(commitMessageFilePath).isFile() &&
+    !!fs.readFileSync(commitMessageFilePath, 'utf8')
+  ) {
+    console.warn('A commit message is already set, cancelling gitmoji\n')
+    process.exit(0)
+  }
+}
+
 export default withHook
--- a/src/commands/commit/index.js
+++ b/src/commands/commit/index.js
@@ -3,11 +3,17 @@ import inquirer from 'inquirer'
 
 import getEmojis from '../../utils/getEmojis'
 import prompts from './prompts'
-import withHook, { registerHookInterruptionHandler } from './withHook'
+import withHook, {
+  registerHookInterruptionHandler,
+  cancelIfMessageIsAlreadySet
+} from './withHook'
 import withClient from './withClient'
 
 const commit = (mode: 'client' | 'hook') => {
-  if (mode === 'hook') registerHookInterruptionHandler()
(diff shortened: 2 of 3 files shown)


Jeremie-Chauvel:
Hi @carloscuesta,
do you think you could take a look ?

carloscuesta:
Hey @Jeremie-Chauvel 

Sure I'm taking a look right now. Sorry for the delay I was on vacation

Thanks!

src/commands/commit/withHook/index.js
```diff
@@ -26,4 +26,15 @@ export const registerHookInterruptionHandler = () => {
   })
 }
 
+export const cancelIfMessageIsAlreadySet = () => {
+  if (
+    fs.existsSync(process.argv[3]) &&
+    fs.lstatSync(process.argv[3]).isFile() &&
+    !!fs.readFileSync(process.argv[3], 'utf8')
```
carloscuesta, line 33: Do you think we can wrap the `process.argv[3]` on a variable and give it a explanation name ?

The name should relate why we use `3` from the `process.argv`

Jeremie-Chauvel: opted for a basic `commitMessageFilePath`

carloscuesta: That works perfect!

carloscuesta:
Hey @Jeremie-Chauvel `3.2.8` which includes your change has been released and published to NPM!

PR #433, opened 2020-08-09
Add options coAuthors (with contacts) and issue/pr reference to commit command

sfelix-martins:
## Description

Add options `--coAuthors` and `--refs` options to allow add co-authors and issue/pr. 
Add still new configuration to add co-authors contacts, for frequent co-authors .

Now with the following command `gitmoji -c --coAuthors --refs` (both optional) the user can define Co-authors and Issue / PR reference. E.g.:

Screen-Recording-2020-08-09-at-19 10 18

## Tests

- [X] All tests passed.
- [X] New features tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,10 @@ A gitmoji interactive client for using gitmojis on commit messages.
   Options
     --init, -i      Initialize gitmoji as a commit hook
     --remove, -r    Remove a previously initialized commit hook
-    --config, -g    Setup gitmoji-cli preferences.
+    --config, -g     Setup gitmoji-cli preferences.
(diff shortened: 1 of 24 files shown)


carloscuesta:
Hey @sfelix-martins 

Thanks for your PR!

However this feature is already covered by the `message` option.

There you can introduce issue PR references and Co-Authors, so I don't think we strictly need this as another question!

sfelix-martins:
@carloscuesta , Thank you for this great package.

So, I did not have a good experience adding co-authors with a single message option. The co-authors contacts improve a lot the usability in teams that practices pair programming.

But, I understood your point.

PR #441, opened 2020-08-14
Add brew as an install method

jevangelista-quizlet:
## Description

Mention the option to install gitmoji via brew, which is easier if you are already using it

Issue: N/A

## Tests
No code change

- [x] All tests passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,13 @@ This project provides an easy solution for using [**gitmoji**](https://github.co
 
 ## Install
 
+### npm
+
 ```bash
 $ npm i -g gitmoji-cli
 ```
 
-or, on macOS, using brew
+### brew
 
 ```bash
 $ brew install gitmoji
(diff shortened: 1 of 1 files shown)


carloscuesta:
Hey! @jevangelista-quizlet

This question may sound silly but how is that even possible? As far as I know I’m not publishing gitmoji-cli to the brew repository, so who is publishing the binaries to brew?

I would like to know this before merging this PR since i want to ensure the source of that package

carloscuesta:
I see this was introduced at https://github.com/Homebrew/homebrew-core/pull/33660

jevangelista-quizlet:
> Hey! @jevangelista-quizlet
> 
> This question may sound silly but how is that even possible? As far as I know I’m not publishing gitmoji-cli to the brew repository, so who is publishing the binaries to brew?
> 
> I would like to know this before merging this PR since i want to ensure the source of that package

Hi! thank you for implementing this in the first place. I love the idea!

TBH I ran `brew search gitmoji` just out of curiosity 🙂

Issue #444, opened 2020-08-18
Commit hook is not working anymore since 3.2.9 upgrade


vanwalj:
Hello @carloscuesta!

Issue:
whenever I `git commit` I get this error message: `A commit message is already set, cancelling gitmoji`

I was using Gitmoji through the git commit hook, setup with gitmoji -i and everything was working fine but it stoped working after upgrading to 3.2.9

**OS**: macOS 10.15.6 (19G2021)
**gitmoji -v**: 3.2.9 installed via brew
**Node version**: 14.8.0
**Git version**: 2.28.0

**Error Message**:
A commit message is already set, cancelling gitmoji

Thanks!

carloscuesta:
That's probably related with the change we introduced at https://github.com/carloscuesta/gitmoji-cli/pull/415 @Jeremie-Chauvel could you take a look at it ? :pray:

oakreef:
I am also experiencing this this morning. I have no other hooks set up or anything else that would be setting a commit message and when I go to commit it opens up my editor with a blank commit message. Nevertheless gitmoji is displaying the line `A commit message is already set, cancelling gitmoji`

**OS:** Ubuntu 20.04
**gitmoji:** 3.2.9
**Node:** 10.19.0
**Git:** 2.25.1

gitmoji installed via npm

carloscuesta:
Yes I reproduced the error, I'm going to revert #415 and release a new version until we found a fix for that

vanwalj:
Awesome, thanks for your work btw!

carloscuesta:
Sorry for the inconvenience 😅 

This should been fixed on `3.2.10` which is published right now.

@Jeremie-Chauvel I reverted #415. If you could take a look to fix this regression we could integrate it again on the cli.

Jeremie-Chauvel:
Hi @carloscuesta 
Sorry for the inconvenience, I will look into it.
Might need to drop the current way of detecting no interactive (based on presence of git message) and instead use `git status` or some less clever way of detecting that the hook shouldn't be launched.

PR #445, opened 2020-08-21
✨ Cancel hook on rebase

Jeremie-Chauvel:
## Description
Hi @carloscuesta, I recreated the feature to cancel the hook on rebase.

Noteworthy point, I contacted @adriencaccia, which implemented a elegant solution on a fork to prefill the commit message in this case, he will come back to you with a proposition .

Issue: https://github.com/carloscuesta/gitmoji-cli/issues/444

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/commit/index.js
+++ b/src/commands/commit/index.js
@@ -3,15 +3,25 @@ import inquirer from 'inquirer'
 
 import getEmojis from '../../utils/getEmojis'
 import prompts from './prompts'
-import withHook, { registerHookInterruptionHandler } from './withHook'
+import withHook, {
+  registerHookInterruptionHandler,
+  cancelIfRebasing
+} from './withHook'
 import withClient from './withClient'
 
 export type CommitMode = 'client' | 'hook'
 
 const commit = (mode: CommitMode) => {
-  if (mode === 'hook') registerHookInterruptionHandler()
+  if (mode === 'hook') {
+    registerHookInterruptionHandler()
+    return cancelIfRebasing().then(() => promptAndCommit(mode))
+  }
 
-  return getEmojis()
+  return promptAndCommit(mode)
+}
+
+const promptAndCommit = (mode: 'client' | 'hook') =>
+  getEmojis()
     .then((gitmojis) => prompts(gitmojis, mode))
     .then((questions) => {
       inquirer.prompt(questions).then((answers) => {
@@ -20,6 +30,5 @@ const commit = (mode: CommitMode) => {
         return withClient(answers)
       })
     })
-}
 
 export default commit
--- a/src/commands/commit/withHook/index.js
+++ b/src/commands/commit/withHook/index.js
@@ -1,4 +1,5 @@
 // @flow
+import execa from 'execa'
 import fs from 'fs'
 
 import { type Answers } from '../prompts'
@@ -25,5 +26,19 @@ export const registerHookInterruptionHandler = () => {
     process.exit(0)
   })
 }
+
+export const cancelIfRebasing = () =>
+  execa('git', ['rev-parse', '--absolute-git-dir']).then(
+    ({ stdout: gitDirectory }) => {
+      // see https://stackoverflow.com/questions/3921409/how-to-know-if-there-is-a-git-rebase-in-progress
+      // to understand how a rebase is detected
+      if (
+        fs.existsSync(gitDirectory + '/rebase-merge') ||
+        fs.existsSync(gitDirectory + '/rebase-apply')
+      ) {
+        process.exit(0)
+      }
+    }
+  )
 
 export default withHook
--- a/test/commands/commit.spec.js
+++ b/test/commands/commit.spec.js
@@ -132,6 +132,8 @@ describe('commit command', () => {
         getDefaultCommitContent.mockReturnValueOnce(
           stubs.emptyDefaultCommitContent
         )
+        execa.mockReturnValueOnce(Promise.resolve(stubs.gitAbsoluteDir))
+        fs.existsSync.mockReturnValueOnce(false).mockReturnValueOnce(false)
         commit('hook')
       })
 
@@ -159,6 +161,8 @@ describe('commit command', () => {
         getDefaultCommitContent.mockReturnValueOnce(
           stubs.emptyDefaultCommitContent
         )
+        execa.mockReturnValueOnce(Promise.resolve(stubs.gitAbsoluteDir))
+        fs.existsSync.mockReturnValueOnce(false).mockReturnValueOnce(false)
         commit('hook')
       })
 
@@ -170,6 +174,28 @@ describe('commit command', () => {
       })
 
       it('should call process.exit', () => {
+        expect(process.exit).toHaveBeenCalledWith(0)
+      })
+    })
+
+    describe('when rebasing', () => {
+      it('should cancel the hook', async () => {
+        /*
+        Use an exception to suspend code execution to simulate the process.exit triggered
+        when the hook mode detect that the user is rebasing. (Simulated to not kill the tests)
+        Simulation needed because if we just mock process exit, then the code execution resume in the test.
+        */
+        mockProcess.mockProcessExit(new Error('ProcessExit0'))
+        execa.mockReturnValueOnce(Promise.resolve(stubs.gitAbsoluteDir))
+        // mock that we found one of the rebase trigger (file existence in .git)
+        fs.existsSync.mockReturnValueOnce(true)
+
+        try {
+          await commit('hook')
+        } catch (e) {
+          expect(e.message).toMatch('ProcessExit0')
+        }
+
         expect(process.exit).toHaveBeenCalledWith(0)
       })
     })
@@ -186,6 +212,9 @@ describe('commit command', () => {
         jest.spyOn(process, 'kill').mockImplementation((pid, signal) => {
           processEvents[signal]()
         })
+
+        execa.mockReturnValueOnce(Promise.resolve(stubs.gitAbsoluteDir))
+        fs.existsSync.mockReturnValueOnce(false).mockReturnValueOnce(false)
 
         inquirer.prompt.mockImplementation(() => {
           process.kill(process.pid, 'SIGINT')


test/setupTests.js
```diff
@@ -12,7 +12,7 @@ jest.mock('ora', () =>
 )
 jest.mock('conf')
 jest.mock('inquirer')
-jest.mock('execa')
+jest.mock('execa', () => jest.fn(() => Promise.resolve({ stdout: 'master' })))
```
carloscuesta, line 15: I would leave this as a general mock like this:

```suggestion
jest.mock('execa')
```

And then mock on the tests as needed the parts of the module you'll need to mock

test/commands/commit.spec.js
```diff
@@ -118,6 +118,7 @@ describe('commit command', () => {
         getEmojis.mockResolvedValue(stubs.gitmojis)
         mockProcess.mockProcessExit()
         process.argv[3] = stubs.argv
+        execa.mockReturnValueOnce(Promise.resolve({ stdout: 'master' }))
```
carloscuesta, line 121: I think we should update the describe to match to the new spec

test/commands/commit.spec.js
```diff
@@ -142,6 +143,7 @@ describe('commit command', () => {
...
```
carloscuesta, line 146: I think we should update the describe to match to the new spec

test/commands/commit.spec.js
```diff
@@ -157,9 +159,30 @@ describe('commit command', () => {
       })
     })
 
+    describe('when rebasing', () => {
+      it('should cancel the hook', async () => {
+        // Use an exception to suspend code execution to simulate process.exit
+        mockProcess.mockProcessExit(new Error('ProcessExit0'))
```
carloscuesta, line 165: Why we can't simulate process.exit here ? 🤔

test/commands/commit.spec.js
```diff
@@ -170,6 +193,8 @@ describe('commit command', () => {
           processEvents[signal]()
         })
 
+        execa.mockReturnValueOnce(Promise.resolve({ stdout: 'master' }))
```
carloscuesta, line 196: We need to update the describe expect to match this new condition here

src/commands/commit/withHook/index.js
```diff
@@ -26,4 +27,13 @@ export const registerHookInterruptionHandler = () => {
   })
 }
 
+export const cancelIfRebasing = (mode: 'hook' | 'client') =>
+  mode !== 'hook'
```
carloscuesta, line 31: Why not making this simpler and then switch the ternary order

```suggestion
  mode === 'hook'
```

src/commands/commit/withHook/index.js
```diff
@@ -26,4 +27,13 @@ export const registerHookInterruptionHandler = () => {
...
+    ? Promise.resolve()
+    : execa('git', ['branch']).then(({ stdout: branchNames }) => {
+        if (/no branch/.test(branchNames)) {
+          process.exit(0)
+        }
+      })
```
carloscuesta, line 37:
```suggestion
    ? execa('git', ['branch']).then(({ stdout: branchNames }) => {
        if (/no branch/.test(branchNames)) {
          process.exit(0)
        }
      })
    : Promise.resolve() 
```

src/commands/commit/withHook/index.js
```diff
@@ -26,4 +27,13 @@ export const registerHookInterruptionHandler = () => {
   })
 }
 
+export const cancelIfRebasing = (mode: 'hook' | 'client') =>
```
carloscuesta, line 30: I think we can make this function `async` to benefit from this and be able to use `await` keyword inside of the fn

src/commands/commit/withHook/index.js
carloscuesta, line 37: Can you explain why you're looking to branch names ? That's why because on the `rebase` process you're not on a named branch but on a commit revision ? There's another way to check if we're on a rebase process ?

src/commands/commit/index.js
```diff
@@ -3,21 +3,28 @@ import inquirer from 'inquirer'
...
+  if (mode === 'hook') {
+    registerHookInterruptionHandler()
+  }
 
-  return getEmojis()
-    .then((gitmojis) => prompts(gitmojis))
-    .then((questions) => {
-      inquirer.prompt(questions).then((answers) => {
-        if (mode === 'hook') return withHook(answers)
+  return cancelIfRebasing(mode).then(() =>
```
carloscuesta, line 17: There's something I think we can improve here.

If we're already making a conditional for the `mode` inside of this function maybe we can reuse it and execute the `cancelIfRebasing` function here

carloscuesta (commented):
Thanks! A small details about the code ❤️

carloscuesta:
I've been testing out this branch locally linked and I found something weird:

When I try to make a commit with the hook option and I finish the prompts. The gitmoji-cli process does not ends until I hit CTRL+C

We should fix this behaviour before merging this:

Demo attached



To test your changes you'll need to uninstall the cli from your system globally:

````
npm uninstall -g gitmoji-cli
```

Clone the project, checkout to your branch and run 

```
npm run build && npm run link
```

Jeremie-Chauvel:
Hi thanks for the feedback,
concerning the hang, I pretty sure I fixed it, I will look into it again.

src/commands/commit/index.js
Jeremie-Chauvel: Sadly, it's not possible to use a function with await in if(mode) since then:
- We need to amke commit an async func
- Then in 'client' mode, we have the 'hang' that you observed in hook mode (the cli doesn't complete until ctrl-C)

src/commands/commit/withHook/index.js
Jeremie-Chauvel: Hi exactly during a rebase, the current branch is `no branch, rebasing <branch name>`, I suppose using git status could do as well

src/commands/commit/withHook/index.js
Jeremie-Chauvel: see https://github.com/carloscuesta/gitmoji-cli/pull/445#discussion_r477316501

src/commands/commit/withHook/index.js
Jeremie-Chauvel: ok for me

src/commands/commit/withHook/index.js
Jeremie-Chauvel: ok for me

test/commands/commit.spec.js
Jeremie-Chauvel: like `while committing on master`?

test/commands/commit.spec.js
Jeremie-Chauvel: I'm simulating process exit by using an exception, if process exit 1 happen during the test, the test is killed, that's why I mock the process exit. I'm using an exception to interrupt as well the code execution to further mimic a process exit

Jeremie-Chauvel:
Hi @carloscuesta
I addressed your feedback:
- I found a way to use the branch mode for this new is rebase test
- I also used a different way to detect rebase, using a git folder created during a rebase see https://stackoverflow.com/questions/3921409/how-to-know-if-there-is-a-git-rebase-in-progress for more details.

## client mode:


## hook out of rebase:


## hook during rebase:

Jeremie-Chauvel:
hmm, concerning you freeze, it's caused by using client mode with hook registred, I'm looking into it, as I can reproduce it

Jeremie-Chauvel:
so, I added a fix for this as well, I tracked it down to a missing encoding in `isHookCreated

`

Jeremie-Chauvel:
@carloscuesta could you take a look when you have some free time ?
Would be much appreciated, thanks.

carloscuesta:
Sure I will take a look at the PR <3

carloscuesta:
Hello @Jeremie-Chauvel 

I found the freeze cause and it's not related with your changes!

I'm going to put up a PR to fix it. It something introduced in another PR.

Afrer that I will review the PR

Thanks for the patience ❤️

carloscuesta:
This PR https://github.com/carloscuesta/gitmoji-cli/pull/461 fixes the "freeze" thing

src/utils/isHookCreated.js
```diff
@@ -11,7 +11,7 @@ const isHookCreated = async () => {
     if (!fs.existsSync(stdout + HOOK.PATH)) {
       return false
     }
-    return fs.readFileSync(stdout + HOOK.PATH) === HOOK.CONTENTS
+    return fs.readFileSync(stdout + HOOK.PATH, 'utf8') === HOOK.CONTENTS
```
carloscuesta, line 14: I fixed this aswell! So when you rebase it there's no need to change that anymore! 🎉

src/commands/commit/index.js
```diff
@@ -3,13 +3,23 @@ import inquirer from 'inquirer'
...
+  if (mode === 'hook') {
+    registerHookInterruptionHandler()
+    return cancelIfRebasing().then(() => promptAndCommit(mode))
+  }
 
-  return getEmojis()
+  return promptAndCommit(mode)
+}
+
+const promptAndCommit = (mode: 'client' | 'hook') =>
```
carloscuesta, line 21: Can you create a `type CommitMode` for the `client` | `hook` thing ? Then we wont have to repeat those values!

test/utils/isHookCreated.spec.js
```diff
@@ -27,7 +27,8 @@ describe('isHookCreated', () => {
     fs.existsSync.mockReturnValue(true)
     expect(await isHookCreated()).toEqual(false)
     expect(fs.readFileSync).toHaveBeenCalledWith(
-      stubs.gitAbsoluteDir + HOOK.PATH
+      stubs.gitAbsoluteDir + HOOK.PATH,
+      'utf8'
```
carloscuesta, line 31: Not needed anymore!

test/commands/commit.spec.js
```diff
@@ -142,6 +144,8 @@ describe('commit command', () => {
         getEmojis.mockResolvedValue(stubs.gitmojis)
         mockProcess.mockProcessExit()
         process.argv[3] = stubs.argv
+        execa.mockReturnValueOnce(Promise.resolve(stubs.gitAbsoluteDir))
+        fs.existsSync.mockReturnValueOnce(false).mockReturnValueOnce(false)
```
carloscuesta, line 148: Why we need to add those mocks inside this test ?

test/commands/commit.spec.js
```diff
@@ -118,6 +118,8 @@ describe('commit command', () => {
...
```
carloscuesta, line 122: Why we need to add those mocks inside this test ?

test/commands/commit.spec.js
```diff
@@ -170,6 +191,9 @@ describe('commit command', () => {
           processEvents[signal]()
         })
 
+        execa.mockReturnValueOnce(Promise.resolve(stubs.gitAbsoluteDir))
+        fs.existsSync.mockReturnValueOnce(false).mockReturnValueOnce(false)
```
carloscuesta, line 195: Why we need to add those mocks inside this test ?

test/commands/commit.spec.js
```diff
@@ -157,9 +161,26 @@ describe('commit command', () => {
       })
     })
 
+    describe('when rebasing', () => {
+      it('should cancel the hook', async () => {
+        // Use an exception to suspend code execution to simulate process.exit
+        mockProcess.mockProcessExit(new Error('ProcessExit0'))
```
carloscuesta, line 167: It would be cool to "simulate" the "rebase" to test the `cancelIfRebasing` code 😊

carloscuesta (commented):
I really like this solution!

Is much more simple and easy to follow ❤️ 

Thanks for the work you're putting in this PR we are really close to ship it!

test/commands/commit.spec.js
Jeremie-Chauvel: that's more or less what I'm doing when I mock ` fs.existsSync` to simulate detection of rebase

test/commands/commit.spec.js
Jeremie-Chauvel: to pass the is `cancelIfRebasing` test

test/commands/commit.spec.js
Jeremie-Chauvel: to pass the is `cancelIfRebasing` test

test/commands/commit.spec.js
Jeremie-Chauvel: to pass the is `cancelIfRebasing` test

carloscuesta (approved):
Let's ship this!

Thanks 🎉

PR #446, opened 2020-08-21
✨ Prefill title and message in hook mode

adriencaccia:
## Description

Hi @carloscuesta, this PR adds prefilling the `title` and `message` prompts when in `hook` mode.
This allows to keep your titles and messages in your history.

gitmoji-prefill-hook

### DIFF
--- a/src/commands/commit/prompts.js
+++ b/src/commands/commit/prompts.js
@@ -3,6 +3,8 @@ import inquirer from 'inquirer'
 
 import configurationVault from '../../utils/configurationVault'
 import filterGitmojis from '../../utils/filterGitmojis'
+import getDefaultCommitContent from '../../utils/getDefaultCommitContent'
+import { type CommitMode } from './index'
 import guard from './guard'
 
 const TITLE_MAX_LENGTH_COUNT: number = 48
@@ -23,40 +25,48 @@ export type Answers = {
   message: string
 }
 
-export default (gitmojis: Array<Gitmoji>): Array<Object> => [
-  {
-    name: 'gitmoji',
-    message: 'Choose a gitmoji:',
-    type: 'autocomplete',
-    source: (answersSoFor: any, input: string) => {
-      return Promise.resolve(
-        filterGitmojis(input, gitmojis).map((gitmoji) => ({
-          name: `${gitmoji.emoji}  - ${gitmoji.description}`,
-          value: gitmoji[configurationVault.getEmojiFormat()]
-        }))
-      )
+export default (gitmojis: Array<Gitmoji>, mode: CommitMode): Array<Object> => {
+  const { title, message } = getDefaultCommitContent(mode)
+
+  return [
+    {
+      name: 'gitmoji',
+      message: 'Choose a gitmoji:',
+      type: 'autocomplete',
+      source: (answersSoFor: any, input: string) => {
+        return Promise.resolve(
+          filterGitmojis(input, gitmojis).map((gitmoji) => ({
+            name: `${gitmoji.emoji}  - ${gitmoji.description}`,
+            value: gitmoji[configurationVault.getEmojiFormat()]
+          }))
+        )
+      }
+    },
+    ...(configurationVault.getScopePrompt()
+      ? [
+          {
+            name: 'scope',
+            message: 'Enter the scope of current changes:',
+            validate: guard.scope
+          }
+        ]
+      : []),
+    {
+      name: 'title',
+      message: 'Enter the commit title',
+      validate: guard.title,
+      transformer: (input: string) => {
+        return `[${
+          (title || input).length
+        }/${TITLE_MAX_LENGTH_COUNT}]: ${input}`
+      },
+      ...(title ? { default: title } : {})
+    },
+    {
+      name: 'message',
+      message: 'Enter the commit message:',
+      validate: guard.message,
+      ...(message ? { default: message } : {})
     }
-  },
-  ...(configurationVault.getScopePrompt()
-    ? [
-        {
-          name: 'scope',
-          message: 'Enter the scope of current changes:',
-          validate: guard.scope
-        }
-      ]
-    : []),
-  {
-    name: 'title',
-    message: 'Enter the commit title',
-    validate: guard.title,
-    transformer: (input: string) => {
-      return `[${input.length}/${TITLE_MAX_LENGTH_COUNT}]: ${input}`
-    }
-  },
-  {
-    name: 'message',
-    message: 'Enter the commit message:',
-    validate: guard.message
-  }
-]
+  ]
+}
--- a/src/utils/getDefaultCommitContent.js
+++ b/src/utils/getDefaultCommitContent.js
@@ -1,0 +1,41 @@
+// @flow
+import fs from 'fs'
+
+import { type CommitMode } from '../commands/commit'
+
+const COMMIT_FILE_PATH_INDEX = 3
+const COMMIT_TITLE_LINE_INDEX = 0
+const COMMIT_MESSAGE_LINE_INDEX = 2
+
+const getDefaultCommitContent = (
+  mode: CommitMode
+): { title: ?string, message: ?string } => {
+  /*
+    Since the hook is called with `gitmoji --hook $1`
+    According to https://git-scm.com/docs/githooks#_prepare_commit_msg,
+    the commit file path will be stored in the 4th argument of the command
+  */
+  const commitFilePath: string = process.argv[COMMIT_FILE_PATH_INDEX]
+
+  if (mode === 'client' || !fs.existsSync(commitFilePath)) {
+    return {
+      title: null,
+      message: null
+    }
+  }
+
+  const commitFileContent: Array<string> = fs
+    .readFileSync(commitFilePath)
+    .toString()
+    .split('\n')
+
+  return {
+    title: commitFileContent[COMMIT_TITLE_LINE_INDEX],
+    message:
+      commitFileContent.length > COMMIT_MESSAGE_LINE_INDEX
+        ? commitFileContent[COMMIT_MESSAGE_LINE_INDEX]
+        : null
+  }
+}
+
+export default getDefaultCommitContent
--- a/test/utils/getDefaultCommitContent.spec.js
+++ b/test/utils/getDefaultCommitContent.spec.js
@@ -1,0 +1,40 @@
+import fs from 'fs'
+import getDefaultCommitContent from '../../src/utils/getDefaultCommitContent'
+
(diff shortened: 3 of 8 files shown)


adriencaccia:
@carloscuesta Hello, I tried to have 100% in diff coverage, however it seems that the `prompts.js` file is not counted in the coverage (it might so because it is a snapshot test).

Given that the coverage was already low for this file, would you say that the tests I have written are sufficient ?

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "javascript.validate.enable": false,
+  "typescript.validate.enable": false,
+  "editor.formatOnSave": true
+}
```
carloscuesta, line 5: Let's put this on the `.gitignore` file :pray:

src/commands/commit/prompts.js
```diff
@@ -23,40 +24,51 @@ export type Answers = {
...
-    }
-  },
-  {
-    name: 'message',
-    message: 'Enter the commit message:',
-    validate: guard.message
-  }
-]
+  ]
+}
```
carloscuesta, line 74: Things we can improve here:

- The `client` | `mode` type should be extracted into the `index` file and imported here to not repeat over and over again

```
// @flow
import inquirer from 'inquirer'

import configurationVault from '../../utils/configurationVault'
import filterGitmojis from '../../utils/filterGitmojis'
import getDefaultCommitContent from '../../utils/getDefaultCommitContent'
import { type CommitMode } from './index'
import guard from './guard'

const TITLE_MAX_LENGTH_COUNT: number = 48

inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'))

export type Gitmoji = {
  code: string,
  description: string,
  emoji: string,
  name: string
}

export type Answers = {
  gitmoji: string,
  scope?: string,
  title: string,
  message: string
}

export default (gitmojis: Array<Gitmoji>, mode: CommitMode): Array<Object> => [
  {
    name: 'gitmoji',
    message: 'Choose a gitmoji:',
    type: 'autocomplete',
    source: (answersSoFor: any, input: string) => {
      return Promise.resolve(
        filterGitmojis(input, gitmojis).map((gitmoji) => ({
          name: `${gitmoji.emoji}  - ${gitmoji.description}`,
          value: gitmoji[configurationVault.getEmojiFormat()]
        }))
      )
    }
  },
  ...(configurationVault.getScopePrompt()
    ? [
        {
          name: 'scope',
          message: 'Enter the scope of current changes:',
          validate: guard.scope
        }
      ]
    : []),
  {
    name: 'title',
    validate: guard.title,
      return `[${
        (title || input).length
      }/${TITLE_MAX_LENGTH_COUNT}]: ${input}`
    },
    default: mode === 'hook'
      ? getDefaultCommitContent().title
      : undefined
  },
  {
    name: 'message',
    message: 'Enter the commit message:',
    validate: guard.message,
    default: mode === 'hook'
      ? getDefaultCommitContent().message
      : undefined
  }
]
```

src/utils/getDefaultCommitContent.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import fs from 'fs'
+
+const getDefaultCommitContent = () => {
+  try {
+    const commitFilePath: string = process.argv[3]
```
carloscuesta, line 6: I think we need to clarify or move those magic numbers into constants that explain why we use those values

src/utils/getDefaultCommitContent.js
```diff
@@ -0,0 +1,23 @@
...
+    const commitFileContent: Array<string> = fs
+      .readFileSync(commitFilePath)
+      .toString()
+      .split('\n')
+    return {
+      title: commitFileContent.length ? commitFileContent[0] : '',
+      message: commitFileContent.length >= 3 ? commitFileContent[2] : ''
```
carloscuesta, line 13: I think we need to clarify or move those magic numbers into constants that explain why we use those values

src/utils/getDefaultCommitContent.js
```diff
@@ -0,0 +1,23 @@
...
+    }
+  } catch (e) {
```
carloscuesta, line 15: We're catching the error and not doing anything ? 🤔

test/utils/getDefaultCommitContent.spec.js
```diff
@@ -0,0 +1,35 @@
+import fs from 'fs'
+import getDefaultCommitContent from '../../src/utils/getDefaultCommitContent'
+
+describe('getDefaultCommitContent', () => {
+  it('should retrieve title and message when file exists', () => {
```
carloscuesta, line 5: I would move those it on different describes:

describe(when file exists)
  it(should retreive title and message)

describe(when file does not exists)
  it(should retrieve ...)

carloscuesta (commented):
👀  Some questions and comments!

Thanks for sending a PR ❤️

src/commands/commit/prompts.js
adriencaccia: Ok for the type !
For the `default` properties definition on the other hand, I will have to keep the syntax where I initialize `title` and `message` before returning the array.
Indeed, with a syntax like the one you suggested, when running a simple `git commit` without messages, the user will be shown the following, resulting of `default: undefined`
image

Moreover, with this syntax, the `getDefaultCommitContent` will be called only once

carloscuesta: Yes! Makes sense to keep then the variables on the function 😊

adriencaccia:
I implemented the changes you requested, and replaced magic numbers in `getDefaultCommitContent.js` with a comment and some constants to better explain their roles.

test/utils/getDefaultCommitContent.spec.js
```diff
@@ -0,0 +1,34 @@
+import fs from 'fs'
+import getDefaultCommitContent from '../../src/utils/getDefaultCommitContent'
+
+describe('getDefaultCommitContent', () => {
+  describe('when file exists', () => {
+    beforeAll(() => {
+      process.argv[3] = 'commit'
```
carloscuesta, line 7: It's not dangerous to overwrite the `process.argv` inside of tests? 🤔

test/utils/getDefaultCommitContent.spec.js
```diff
@@ -0,0 +1,34 @@
...
+      const { title, message } = getDefaultCommitContent()
+
+      expect(title).toBe('commit title')
+      expect(message).toBe('commit message')
+    })
+  })
+
+  describe('when file does not exist', () => {
+    beforeAll(() => {
+      process.argv[3] = 'commit'
```
carloscuesta, line 23: It's not dangerous to overwrite the `process.argv` inside of tests? 🤔

src/commands/commit/prompts.js
```diff
@@ -23,40 +25,50 @@ export type Answers = {
...
-  {
-    name: 'title',
-    message: 'Enter the commit title',
-    validate: guard.title,
-    transformer: (input: string) => {
-      return `[${input.length}/${TITLE_MAX_LENGTH_COUNT}]: ${input}`
+export default (gitmojis: Array<Gitmoji>, mode: CommitMode): Array<Object> => {
+  const { title, message } = mode === 'hook'
+    ? getDefaultCommitContent()
+    : { title: undefined, message: undefined }
```
carloscuesta, line 31: I have a question right here, If we are returning `{ title: undefined, message: undefined }` in here, what's the point then of returning it also inside of the `getDefaultCommitContent` function ?

I mean can we find another way to remove the default values for title and message prompts where the mode it's not hook ?

src/utils/getDefaultCommitContent.js
```diff
@@ -0,0 +1,36 @@
...
+    the commit file path will be stored in the 4th argument of the command
+  */
+  const commitFilePath: string = process.argv[COMMIT_FILE_PATH_INDEX]
+
+  if (fs.existsSync(commitFilePath) === false) {
+    return {
+      title: undefined,
+      message: undefined
+    }
+  }
```
carloscuesta, line 21: This is for when the hook command it's not autocompleted right ? Can we use for example `null` instead of `undefined` in here to remove the default values for inquirer prompt ?

I think we should use `null` instead of `undefined`

src/utils/getDefaultCommitContent.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+import fs from 'fs'
+
+const COMMIT_FILE_PATH_INDEX = 3
+const COMMIT_TITLE_LINE_INDEX = 0
+const COMMIT_MESSAGE_LINE_INDEX = 2
+
+const getDefaultCommitContent = () => {
```
carloscuesta, line 8: It would be cool to type the return value of this function

```suggestion
const getDefaultCommitContent = (): { title: type, message: type } => {
```

src/utils/getDefaultCommitContent.js
```diff
@@ -0,0 +1,36 @@
...
+  /*
+    Since the hook is called with `gitmoji --hook $1`
+    According to https://git-scm.com/docs/githooks#_prepare_commit_msg,
+    the commit file path will be stored in the 4th argument of the command
+  */
+  const commitFilePath: string = process.argv[COMMIT_FILE_PATH_INDEX]
+
+  if (fs.existsSync(commitFilePath) === false) {
```
carloscuesta, line 16:
```suggestion
  if (!fs.existsSync(commitFilePath)) {
```

carloscuesta (commented):
👏🏼

src/commands/commit/prompts.js
adriencaccia: I added the `mode` parameter to `getDefaultCommitContent` util so that we can return null values directly in the function when `mode === 'client'`

src/utils/getDefaultCommitContent.js
adriencaccia: Done

src/utils/getDefaultCommitContent.js
adriencaccia: Done

src/utils/getDefaultCommitContent.js
adriencaccia: Done

test/utils/getDefaultCommitContent.spec.js
adriencaccia: Since we mock the `fs` functions that use the value of `process.argv[3]`, I think we can just not bother ovrewriting it at all

carloscuesta (approved):
I think we're ready to merge this 🚀 

Thanks for the work you've put on this PR @adriencaccia ❤️

Larandar:
Thank you ❤️

PR #464, opened 2020-09-28
Add quotes around commit title and message

scitronboy:
## Description

Commit message and title should be quoted, or else commit titles with special characters can throw errors.

## Tests

- [:heavy_check_mark:  ] All tests passed.


scitronboy:
Closing as apparently I don't think this is actually needed.

Issue #465, opened 2020-10-01
[Feature request]: Add --scope option to gitmoji -c command 


sirgalleto:
Hello @carloscuesta! I would like to discuss a feature that came to my mind.

# Inspiration

When I'm working on a feature, my team uses certain naming convention where the ticket number is placed, for example: 

* feature/ticket-01/add-gitmoji
* fix/ticket-99/fix-bad-emoji

For different branches, I'm usually doing more than one commit, and the process of input the ticket-number as _scope of current changes_ is 1. Error prone, 2. Tedious, 3. Boring, 4. Expensive, 5. I just hate to repeat myself. 

# Proposal

Given a branch name with my team's naming convention, I can create a bash script that outputs my ticket number. 

```sh
#!/bin/sh
branch=$(git symbolic-ref HEAD)
IFS='/'
read -a strarr <<< "$branch"
echo "Book Name : ${strarr[3]} "
```

Having the ticket-number in the context I need a way to inject it into the _gitmoji-cli_, where I would like to have two options 

1. A flag in the -c command 

```sh
$ gitmoji -c --scope=$(myScript)
```

2. An environment variable. 

```sh 
GITMOJI_SCOPE=$(myScript) gitmoji -c 
```

## Personal comments

1. By adding options to receive the prompts as params, `gitmoji-cli` can be used in more different ways giving the final user more productivity and extensibility. 
2. There are certain more options to detect the scope from the branch name _for example adding a regex config_ but that comes with a solution focused in branch names, and I think the best solution is making `gitmoji-cli` more extensible. 

What do you think? 

____

I'm willing to develop this option, but first I want to confirm if you're open to add this, and what are your thoughts of the approach and implementation ideas.  

Thanks for this awesome project!

carloscuesta:
Hey @sirgalleto 

Thanks for opening an issue!

This feature already exists! You can enable the scope prompt via the `gitmoji -g` configuration option.

See:

![Screenshot 2020-10-01 at 08 18 37](https://user-images.githubusercontent.com/7629661/94774699-b88f4a00-03be-11eb-8087-4f1b20c1c38b.png)

![Screenshot 2020-10-01 at 08 18 15](https://user-images.githubusercontent.com/7629661/94774697-b75e1d00-03be-11eb-8a19-bfdc58275a1d.png)

sirgalleto:
Right @carloscuesta I'm already using it, and my apologies to not explain myself well. 

[...]

PR #522, opened 2021-01-03
✨ (feat): Added environment variable support 🥓

jayakornk:
## Description

I just quickly tweak some code to make a few environment variables available. The reason that I made this tweak because I won't be able to use hook on my projects otherwise it will overwrite other hook like husky. In my opinion, adding these as an option will increase flexibility and allow more people to use this awesome tool. (I'm open to suggestions in case I missed some point 😉)

### Added Environment Variables
- **SCOPE_PROMPT** (`boolean`): behave same as the **Scope prompt** option in config
- **SCOPE** (`string`): assign _scope_ directly and skip **Scope prompt**
- **SIGNED_COMMIT** (`boolean`): behave same as the **Signed commits** option in config except that this one enable per commit

Example:
Screen Shot 2021-01-04 at 02 28 07

Issue: #465 

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/commit/withClient/index.js
+++ b/src/commands/commit/withClient/index.js
@@ -9,7 +9,8 @@ import { type Answers } from '../prompts'
 
 const withClient = async (answers: Answers) => {
   try {
-    const scope = answers.scope ? `(${answers.scope}): ` : ''
+    let scope = process.env.SCOPE || answers.scope
+    scope = scope ? `(${scope}): ` : ''
     const title = `${answers.gitmoji} ${scope}${answers.title}`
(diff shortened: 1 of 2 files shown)


carloscuesta:
Hey I'm missing a lot of context on those changes, please open an issue first before creating any PR. You may find an existing issue were we are tracking errors related with Husky.

sirgalleto:
Hi again @carloscuesta, I would like to know why you reject this proposal. In my opinion, having this in place opens the tool to be integrable with other external tools that can tweak the parameters before the user input.

carloscuesta:
It's already answered on the message above

PR #533, opened 2021-01-29
:bug: Fix reference error in getEmojis.js

lonelam:
to pass unit test & prevent to throw uncaught exception

## Description

Issue: #

## Tests

- [x] All tests passed.

### DIFF
--- a/test/utils/emojisCache.spec.js
+++ b/test/utils/emojisCache.spec.js
@@ -18,7 +18,11 @@ describe('emojisCache', () => {
   })
 
   it('should match CACHE_PATH', () => {
-    expect(CACHE_PATH).toMatchSnapshot()
+    if (process.platform !== 'win32') {
+      expect(CACHE_PATH).toMatchSnapshot()
+    } else {
+      expect('/Users/Test/.gitmoji/gitmojis.json').toMatchSnapshot()
+    }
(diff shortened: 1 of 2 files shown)


test/utils/emojisCache.spec.js
```diff
@@ -18,7 +18,11 @@ describe('emojisCache', () => {
   })
 
   it('should match CACHE_PATH', () => {
-    expect(CACHE_PATH).toMatchSnapshot()
+    if (process.platform !== 'win32') {
+      expect(CACHE_PATH).toMatchSnapshot()
+    } else {
+      expect('/Users/Test/.gitmoji/gitmojis.json').toMatchSnapshot()
```
carloscuesta, line 24: From a test point of view, this doesn't makes sense at all

It's like doing `expect(1).toMatchSnapshot()` or `expect(1).toEqual(1)`

carloscuesta (commented):
Hey @lonelam thanks for opening a PR

I don't know what kind of bug you're trying to fix here, please open an issue if this is a real bug providing a reproduction case and we'll take a look around fixing it 

Opening a PR like this without context makes it a little hard to review or know what's the reasoning behind the PR itself

Thanks! 😊

Issue #534, opened 2021-01-30
Option to skip title and message prompt ?


momsse:
Hello @carloscuesta!

Thanks for the great job done with gitmoji and this cli ❤️

Is it possible to add support for an option that will skip `title` and `message` prompt when already filled by the hook (thanks to https://github.com/carloscuesta/gitmoji-cli/pull/446) ?

Issue:

Basically when I run the following command:

```bash
commit -m "title" -m "message"
```

Would love to not have to double enter to skip the already filled `title` / `message` ❤️

<img width="655" alt="Capture d’écran 2021-01-30 à 21 10 39" src="https://user-images.githubusercontent.com/1344228/106366987-b4a90f80-633f-11eb-965a-36f53dc68909.png">

**OS**: Mac
**gitmoji -v**: 3.2.18
**Node version**: 12.18.0

carloscuesta:
Hey @momsse, thanks for opening an issue 😊 

I totally understand your concern, whenever you're prefilling those two questions with the `title` and `message` it doesn't makes any sense to ask for them on the cli. 

However, I think that then prompting the cli only for the gitmoji then it's a "waste" of time because you should already decide your gitmoji before your commit message, so I don't really see the point of using the cli in this way. It doesn't adds any real value to the user

We can try to implement the skipping feature but I think that the ☝🏼 problem is something bigger to consider!

What do you think?

momsse:
Thanks for the reply,

> However, I think that then prompting the cli only for the gitmoji then it's a "waste" of time because you should already > decide your gitmoji before your commit message, so I don't really see the point of using the cli in this way. It doesn't adds any real value to the user

If I always remember some emoji by heart (like `:bug:`, `:sparkles:`). Some of them are uncommon enough that before suing this cli I was wasting my team searching in https://gitmoji.dev the correct one to use and that's actually the value (for me) of that cli. Entering the correct keywords for having the correct gitmoji.

Another great feature (for me) would be to dismiss completely the hook when it detect that we already enter the correct gitmoji :-)

But I completely understand that perhaps my usage is not the one expected by this cli.

Thanks you anyway for the great work and the time responding me :-)

carloscuesta:
It makes sense! If we implement this feature then we should be removing the prefill one since the questions are going to be skipped right?

momsse:
> It makes sense! If we implement this feature then we should be removing the prefill one since the questions are going to be skipped right?

[...]

PR #536, opened 2021-02-05
Add support to skip prompting filled info

momsse:
## Description

Add support to an option that will skip asking `gitmoji`/`title`/`message` when they are filled.

Issue: #534 

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/commit/index.js
+++ b/src/commands/commit/index.js
@@ -8,6 +8,8 @@ import withHook, {
   cancelIfNeeded
 } from './withHook'
 import withClient from './withClient'
+import getDefaultCommitContent from '../../utils/getDefaultCommitContent'
+import getDefaultAnswers from '../../utils/getDefaultAnswers'
 
 export type CommitMode = 'client' | 'hook'
 
@@ -20,15 +22,19 @@ const commit = (mode: CommitMode) => {
   return promptAndCommit(mode)
 }
 
-const promptAndCommit = (mode: 'client' | 'hook') =>
-  getEmojis()
-    .then((gitmojis) => prompts(gitmojis, mode))
-    .then((questions) => {
-      inquirer.prompt(questions).then((answers) => {
-        if (mode === 'hook') return withHook(answers)
+const promptAndCommit = async (mode: CommitMode) => {
+  const gitmojis = await getEmojis()
+  const commitContent = getDefaultCommitContent(mode)
+  const defaultAnswers = getDefaultAnswers(commitContent)
+  const questions = prompts(gitmojis, commitContent, defaultAnswers)
+  const answers = {
+    ...defaultAnswers,
+    ...(await inquirer.prompt(questions))
+  }
 
-        return withClient(answers)
-      })
-    })
+  if (mode === 'hook') return withHook(answers)
+
+  return withClient(answers)
+}
 
 export default commit
--- a/src/commands/commit/prompts.js
+++ b/src/commands/commit/prompts.js
@@ -3,9 +3,8 @@ import inquirer from 'inquirer'
 
 import configurationVault from '../../utils/configurationVault'
 import filterGitmojis from '../../utils/filterGitmojis'
-import getDefaultCommitContent from '../../utils/getDefaultCommitContent'
-import { type CommitMode } from './index'
 import guard from './guard'
+import type { CommitContent } from '../../utils/getDefaultCommitContent'
 
 const TITLE_MAX_LENGTH_COUNT: number = 48
 
@@ -25,11 +24,15 @@ export type Answers = {
   message: string
 }
 
-export default (gitmojis: Array<Gitmoji>, mode: CommitMode): Array<Object> => {
-  const { title, message } = getDefaultCommitContent(mode)
+export default (
+  gitmojis: Array<Gitmoji>,
+  { title, message }: CommitContent,
+  defaultAnswers: $Shape<Answers>
+): Array<Object> => {
+  const questions = []
 
-  return [
-    {
+  if (!defaultAnswers.gitmoji) {
+    questions.push({
       name: 'gitmoji',
       message: 'Choose a gitmoji:',
       type: 'autocomplete',
@@ -41,17 +44,17 @@ export default (gitmojis: Array<Gitmoji>, mode: CommitMode): Array<Object> => {
           }))
         )
       }
-    },
-    ...(configurationVault.getScopePrompt()
-      ? [
-          {
-            name: 'scope',
-            message: 'Enter the scope of current changes:',
-            validate: guard.scope
-          }
-        ]
-      : []),
-    {
+    })
+  }
+  if (configurationVault.getScopePrompt()) {
+    questions.push({
+      name: 'scope',
+      message: 'Enter the scope of current changes:',
+      validate: guard.scope
+    })
+  }
+  if (!defaultAnswers.title) {
+    questions.push({
       name: 'title',
       message: 'Enter the commit title',
       validate: guard.title,
@@ -61,12 +64,17 @@ export default (gitmojis: Array<Gitmoji>, mode: CommitMode): Array<Object> => {
         }/${TITLE_MAX_LENGTH_COUNT}]: ${input}`
       },
       ...(title ? { default: title } : {})
-    },
-    {
+    })
+  }
+
+  if (!defaultAnswers.message) {
+    questions.push({
       name: 'message',
       message: 'Enter the commit message:',
       validate: guard.message,
       ...(message ? { default: message } : {})
-    }
-  ]
+    })
+  }
+
+  return questions
 }
--- a/src/commands/config/prompts.js
+++ b/src/commands/config/prompts.js
@@ -5,7 +5,8 @@ export const CONFIGURATION_PROMPT_NAMES = {
   AUTO_ADD: 'autoAdd',
   EMOJI_FORMAT: 'emojiFormat',
   SCOPE_PROMPT: 'scopePrompt',
-  SIGNED_COMMIT: 'signedCommit'
+  SIGNED_COMMIT: 'signedCommit',
+  SKIP_PROMPTING_FILLED_INFO: 'skipPromptingFilledInfo'
 }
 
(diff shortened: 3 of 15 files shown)


src/commands/commit/prompts.js
```diff
@@ -25,11 +24,15 @@ export type Answers = {
...
+export default (
+  gitmojis: Array<Gitmoji>,
+  { title, message }: CommitContent,
+  defaultAnswers: $Shape<Answers>
+): Array<Object> => {
+  const questions = []
 
-  return [
-    {
+  if (!defaultAnswers.gitmoji) {
```
carloscuesta, line 34: Can we stick to the old format? I just want to keep immutability in the codebase as much as possible 🙏🏼

src/commands/config/prompts.js
```diff
@@ -41,5 +42,11 @@ export default () => [
     message: 'Enable scope prompt',
     type: 'confirm',
     default: configurationVault.getScopePrompt()
+  },
+  {
+    name: CONFIGURATION_PROMPT_NAMES.SKIP_PROMPTING_FILLED_INFO,
+    message: 'Skip prompting already filled info',
```
carloscuesta, line 48: I think we need to clarify better what this option is doing. For you this might be already familiar but for someone who uses the cli and doesn't looks into the PR issues I don't think `Skip prompting already filled info` is an understandable message, what information is referring this to ?

src/commands/config/prompts.js
```diff
@@ -5,7 +5,8 @@ export const CONFIGURATION_PROMPT_NAMES = {
   AUTO_ADD: 'autoAdd',
   EMOJI_FORMAT: 'emojiFormat',
   SCOPE_PROMPT: 'scopePrompt',
-  SIGNED_COMMIT: 'signedCommit'
+  SIGNED_COMMIT: 'signedCommit',
+  SKIP_PROMPTING_FILLED_INFO: 'skipPromptingFilledInfo'
```
carloscuesta, line 9: Same as for this option, is a little bit ambiguous

src/utils/getDefaultCommitContent.js
```diff
@@ -29,8 +39,20 @@ const getDefaultCommitContent = (
...
+  let gitmoji = null
+  let title = commitFileContent[COMMIT_TITLE_LINE_INDEX]
+
+  const gitmojiRegexResult =
+    configurationVault.getEmojiFormat() === EMOJI_COMMIT_FORMATS.CODE
+      ? gitmojiCodeRegex.exec(title)
+      : gitmojiSymbolRegex.exec(title)
+  if (gitmojiRegexResult && gitmojiRegexResult.length === 3) {
+    ;[, gitmoji, title] = gitmojiRegexResult
+  }
```
carloscuesta, line 51: Overall this is fairly complex, there are some magic numbers can you explain what are you trying to do in this block?

src/utils/getDefaultCommitContent.js
```diff
@@ -2,14 +2,23 @@
...
 
 const COMMIT_FILE_PATH_INDEX = 3
 const COMMIT_TITLE_LINE_INDEX = 0
 const COMMIT_MESSAGE_LINE_INDEX = 2
 
-const getDefaultCommitContent = (
-  mode: CommitMode
-): { title: ?string, message: ?string } => {
+const gitmojiSymbolRegex = /^(\p{Emoji_Presentation})\s*(.*)/gu
+const gitmojiCodeRegex = /^(:[a-zA-Z_]+:)\s*(.*)/gu
```
carloscuesta, line 13: Why do we need this?

src/commands/commit/prompts.js
```diff
@@ -3,9 +3,8 @@ import inquirer from 'inquirer'
 
 import configurationVault from '../../utils/configurationVault'
 import filterGitmojis from '../../utils/filterGitmojis'
-import getDefaultCommitContent from '../../utils/getDefaultCommitContent'
-import { type CommitMode } from './index'
```
carloscuesta, line 7: Why we moved this outside of this module?

src/commands/commit/index.js
```diff
@@ -20,15 +22,19 @@ const commit = (mode: CommitMode) => {
...
-const promptAndCommit = (mode: 'client' | 'hook') =>
-  getEmojis()
-    .then((gitmojis) => prompts(gitmojis, mode))
-    .then((questions) => {
-      inquirer.prompt(questions).then((answers) => {
-        if (mode === 'hook') return withHook(answers)
+const promptAndCommit = async (mode: CommitMode) => {
+  const gitmojis = await getEmojis()
+  const commitContent = getDefaultCommitContent(mode)
+  const defaultAnswers = getDefaultAnswers(commitContent)
```
carloscuesta, line 28: Why we moved this at commit level?

carloscuesta (commented):
Thanks for sending a PR!

Overall code is a little bit complex just pointed out some questions and comments

src/commands/config/prompts.js
momsse: Sure perhaps a more verbose sentence of what it does ?

> Skip prompting for gitmoji, title and message commit when retrieved

src/commands/config/prompts.js
momsse: Right, do you have some suggestion ?

src/utils/getDefaultCommitContent.js
momsse: Basically extracting the gitmoji from the title (`3` is the size of the regex array when it succeed).

I can extract the logic to his own function to make it easier to read. Is it fair enough ?

src/utils/getDefaultCommitContent.js
momsse: To skip the gitmoji prompt I have to detect if the user has already filled it on the title part of the commit. This is the regex to do the job.

src/commands/commit/prompts.js
momsse: Because I need the result of the function call for `getDefaultAnswers` and `prompts`. To avoid call it twice I moved it down to `commit/index`

src/commands/commit/index.js
momsse: Same answer as above.

momsse:
To resume:

1. `getDefaultCommitContent` will also extract gitmoji from the commit message.
2. All information retrieved are used by `getDefaultAnswers` 
3. Those default answers are passed to the `prompts` to avoid asking the user information that we retrieved

src/commands/commit/prompts.js
momsse: Sure

carloscuesta:
I think I've got an idea to simplify the implementation of this! 😊 

Taking a look at the [`Inquirer.js` docs]() you can see there's a `when` option that could be configured for every prompt. What this property does?

```
when: (Function, Boolean) Receive the current user answers hash and should return true or false depending on whether or not this question should be asked. The value can also be a simple boolean.
```

So basically, what we can try, is to add the `when` property to the `title` and `message` prompts conditionally based on the `configurationVault` option. Something like:

```js
{
      name: 'title',
      validate: guard.title,
        return `[${
          (title || input).length
        }/${TITLE_MAX_LENGTH_COUNT}]: ${input}`
      },
      ...(title ? { default: title } : {}),
      ...(configurationVault.skipOption ? { when: true } : {})
    },
    {
      name: 'message',
      message: 'Enter the commit message:',
      validate: guard.message,
      ...(message ? { default: message } : {}),
      ...(configurationVault.skipOption ? { when: true } : {})
    }
```

Then we can unify all the `prompts` logic inside of `prompts.js`

What do you think?

momsse:
Hi @carloscuesta,

Thanks for the suggestion. This can simplify the prompt a little bit (we do not need to read `defaultAnswers`).  Unfortunately when we use this option `{ when: false }` the question is skipped but the default answer too :/ This is not a solution that can help us removing `getDefaultAnswers` function.

Tell me if I missed the point. Here the changes I tried:

```typescript
export default (
  gitmojis: Array<Gitmoji>,
  { gitmoji, title, message }: CommitContent
): Array<Object> => {
  const skipPromptingFilledInfo = configurationVault.getSkipPromptingFilledInfo()

  return [
    {
      name: 'gitmoji',
      message: 'Choose a gitmoji:',
      type: 'autocomplete',
      source: (answersSoFor: any, input: string) => {
        return Promise.resolve(
          filterGitmojis(input, gitmojis).map((gitmoji) => ({
            name: `${gitmoji.emoji}  - ${gitmoji.description}`,
            value: gitmoji[configurationVault.getEmojiFormat()]
          }))
        )
      },
      ...(gitmoji ? { default: gitmoji } : {}),
      ...(skipPromptingFilledInfo && gitmoji ? { when: false } : {}),
    },
    {
      name: 'scope',
      message: 'Enter the scope of current changes:',
      validate: guard.scope,
      ...(!configurationVault.getScopePrompt() ? { when: false } : {}),
    },
    {
      name: 'title',
      validate: guard.title,
        return `[${
          (title || input).length
        }/${TITLE_MAX_LENGTH_COUNT}]: ${input}`
      },
      ...(title ? { default: title } : {}),
      ...(skipPromptingFilledInfo && title && guard.title(title) ? { when: false } : {}),
    },
    {
      name: 'message',
      message: 'Enter the commit message:',
      validate: guard.message,
      ...(message ? { default: message } : {}),
      ...(skipPromptingFilledInfo && message && guard.message(message) ? { when: false } : {}),
    }
  ]
}
```

carloscuesta:
Will take a look at this shortly!

carloscuesta:
Hey @momsse sorry for the delay! 

Since we don't have a simple way to introduce this feature at the moment, I think it would be great to skip at this moment (because it's not a big deal for the UX of the cli). Hope we can revisit this on a future iteration and opportunity

Thanks for the time and effort you put on this PR, I really appreciate it, thanks ❤️

Karsens:
Hi, I saw that this wasn't merged. Very unfortunate because I really need this as I usually don't put so much effort into my git commits and would really like them fast. So I won't use gitmoji for now. :( Hope you can merge this in the future!

Issue #577, opened 2021-04-12
Feature request: read core.hooksPath from git config during `gitmoji -i`


flo-sch:
Hello @carloscuesta!

Issue:

`gitmoji -i` currently creates a `prepare-commit-msg` file under the hardcoded `.git/hooks` path, which is the default path for git hooks.
However, this path can be configured `git config --set core.hooksPath path/to/hooks/folder` and this is what for instance [husky v6 started to do](https://github.com/typicode/husky/issues/943).

Would it be possible to create that file at the local config path instead, accessible by running `git config --get core.hooksPath`? (using  [child_process.spawnSync](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options) or such perhaps?)

**OS**: Mac OS 11.2.3
**gitmoji -v**: 3.3.1
**Node version**: 14.16.1

carloscuesta:
Yes! This is much better than assuming the local default hooks directory

Will try to take a look at this but if you want to help feel free!

flo-sch:
I can give it a quick try this week I think

carloscuesta:
Thanks for the help @flo-sch

rushilsrivastava:
@carloscuesta Was #578 bundled with v3.4.0? I looked through the package posted on NPM (see the hook file on [unpkg](https://unpkg.com/browse/gitmoji-cli@3.4.0/lib/commands/hook/hook.js)) and I don't think #578 has been merged in.

The release notes mention that this was intended as well.

rushilsrivastava:
The release published by Github actions seems to be working fine, I think the version published to NPM and Brew have a different version uploaded.

carloscuesta:
That's really weird 🤔 Going to republish now

carloscuesta:
Republished and fixed at 3.4.1 🎉

carloscuesta:
Turns out I forgot to execute the `build` script before publishing 😅. I automated this on a PR to never forgot it again

PR #578, opened 2021-04-14
:sparkles: update hooks create & delete path

flo-sch:
- read core.hooksPath from git config (if set)
- default to "process.cwd()/.git/hooks"
- adjust unit tests

## Description

Change how the path for git hooks is retrieved:
It now invokes `git config --get core.hooksPath` which provides a *relative* path (default to `.git/hooks` if unset), then create/delete the hooks at that path.

NOTE: @carloscuesta, the previous `git rev-parse` command returned an absolute folder.
Not sure if it is worth defaulting to that command, in case `core.hooksPath` is not set? (to be fully 100% backward-compatible?)

(I do not know if people would try to set-up gitmoji outside of the current working directory, feels edge-case to me 🤷‍♂️)

Issue: #577

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/hook/create/index.js
+++ b/src/commands/hook/create/index.js
@@ -11,15 +11,8 @@ const createHook = async () => {
   try {
     const hookFile = await getAbsoluteHooksPath(HOOK.FILENAME)
 
-    fs.writeFile(
-      hookFile,
-      HOOK.CONTENTS,
-      { mode: HOOK.PERMISSIONS },
-      (error) => {
-        if (error) return spinner.fail(error)
-        spinner.succeed('Gitmoji commit hook created successfully')
-      }
-    )
+    fs.writeFileSync(hookFile, HOOK.CONTENTS, { mode: HOOK.PERMISSIONS })
+    spinner.succeed('Gitmoji commit hook created successfully')
   } catch (error) {
     spinner.fail(`Error: ${error}`)
   }
--- a/src/commands/hook/create/index.js
+++ b/src/commands/hook/create/index.js
@@ -1,28 +1,15 @@
 // @flow
 import fs from 'fs'
-import execa from 'execa'
 import ora from 'ora'
-import path from 'path'
 
 import HOOK from '../hook'
+import getAbsoluteHooksPath from '../../../utils/getAbsoluteHooksPath'
 
-/**
- * Create a "prepare-commit-msg" git hook
- *
- * If "core.hooksPath" is not set in git config, use default ".git/hooks"
- * @see https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
- */
 const createHook = async () => {
   const spinner = ora('Creating the gitmoji commit hook').start()
 
   try {
-    const { stdout } = await execa('git', ['config', '--get', 'core.hooksPath'])
-    const relativeHooksPath = stdout || '.git/hooks'
-    const hookFile = path.resolve(
-      process.cwd(),
-      relativeHooksPath,
-      HOOK.FILENAME
-    )
+    const hookFile = await getAbsoluteHooksPath(HOOK.FILENAME)
 
     fs.writeFile(
       hookFile,
--- a/src/commands/hook/create/index.js
+++ b/src/commands/hook/create/index.js
@@ -2,17 +2,30 @@
 import fs from 'fs'
 import execa from 'execa'
 import ora from 'ora'
+import path from 'path'
 
 import HOOK from '../hook'
 
+/**
+ * Create a "prepare-commit-msg" git hook
+ *
+ * If "core.hooksPath" is not set in git config, use default ".git/hooks"
+ * @see https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
+ */
 const createHook = async () => {
   const spinner = ora('Creating the gitmoji commit hook').start()
 
   try {
-    const { stdout } = await execa('git', ['rev-parse', '--absolute-git-dir'])
+    const { stdout } = await execa('git', ['config', '--get', 'core.hooksPath'])
+    const relativeHooksPath = stdout || '.git/hooks'
+    const hookFile = path.resolve(
+      process.cwd(),
+      relativeHooksPath,
+      HOOK.FILENAME
+    )
 
     fs.writeFile(
-      stdout + HOOK.PATH,
+      hookFile,
       HOOK.CONTENTS,
       { mode: HOOK.PERMISSIONS },
       (error) => {
--- a/src/commands/hook/remove/index.js
+++ b/src/commands/hook/remove/index.js
@@ -11,13 +11,10 @@ const removeHook = async () => {
   try {
     const hookFile = await getAbsoluteHooksPath(HOOK.FILENAME)
 
-    fs.unlink(hookFile, (error) => {
-      if (error)
-        return spinner.fail('Error: Gitmoji commit hook is not created')
-      spinner.succeed('Gitmoji commit hook removed successfully')
-    })
+    fs.unlinkSync(hookFile)
+    spinner.succeed('Gitmoji commit hook removed successfully')
   } catch (error) {
-    spinner.fail(`Error: ${error}`)
+    spinner.fail('Error: Gitmoji commit hook is not created')
   }
 }
 
--- a/src/commands/hook/remove/index.js
+++ b/src/commands/hook/remove/index.js
@@ -1,28 +1,15 @@
 // @flow
 import fs from 'fs'
-import execa from 'execa'
 import ora from 'ora'
-import path from 'path'
 
 import HOOK from '../hook'
+import getAbsoluteHooksPath from '../../../utils/getAbsoluteHooksPath'
 
-/**
- * Delete a "prepare-commit-msg" git hook
- *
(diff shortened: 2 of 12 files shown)


src/utils/isHookCreated.js
```diff
@@ -1,13 +1,14 @@
...
 import fs from 'fs'
+import path from 'path'
 
 import HOOK from '../commands/hook/hook'
 
 const isHookCreated = async () => {
   try {
-    const { stdout } = await execa('git', ['rev-parse', '--absolute-git-dir'])
-    const hookFile = stdout + HOOK.PATH
+    const { stdout } = await execa('git', ['config', '--get', 'core.hooksPath'])
```
carloscuesta, line 10: Shouldn't we check here for the default hooks path if no core.hooksPath is provided?

carloscuesta (commented):
Thanks for opening up a PR ❤️ 

Great work! A couple of details:

src/commands/hook/create/index.js
```diff
@@ -2,17 +2,30 @@
...
+ * If "core.hooksPath" is not set in git config, use default ".git/hooks"
+ * @see https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
+ */
 const createHook = async () => {
   const spinner = ora('Creating the gitmoji commit hook').start()
 
   try {
-    const { stdout } = await execa('git', ['rev-parse', '--absolute-git-dir'])
+    const { stdout } = await execa('git', ['config', '--get', 'core.hooksPath'])
+    const relativeHooksPath = stdout || '.git/hooks'
```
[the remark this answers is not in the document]
flo-sch: That would potentially get me:

1. `/Users/foo/projects/bar/.git`
2. `.git/hooks`

Should I then merge those by replacing the common `.git` do you think?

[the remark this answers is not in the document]
flo-sch: Nevermind, according to the doc, the `core.hooksPath` in config could be either absolute or relative actually...
I will come with a new version hopefully covering all cases!

flo-sch:
Alright, I forced pushed after signing my commits, I use 2 emails for my Git accounts and I often forget to switch to my personal one when contributing to OSS 😅 

Anyway, I have created a new helper to retrieve the hooks path following different scenarios:

1. If the hook path in the config is absolute, put the hook in there (apparently possible according to git documentation)
ex: `/etc/git/hooks`
2. If it is relative, get the absolute path of the git directory, use `dirname()` to remove the `.git` part, then apply the hooks path (since according to my understanding, it is relative to the root folder)
ex: `/Users/foo/projects/bar/.git/hooks` or `/Users/carloscuesta/GitHub/gitmoji-cli/.husky`

src/commands/hook/remove/index.js
```diff
@@ -1,23 +1,20 @@
...
-    const { stdout } = await execa('git', ['rev-parse', '--absolute-git-dir'])
+    const hookFile = await getAbsoluteHooksPath(HOOK.FILENAME)
 
-    fs.unlink(stdout + HOOK.PATH, (error) => {
-      if (error)
-        return spinner.fail('Error: Gitmoji commit hook is not created')
-      spinner.succeed('Gitmoji commit hook removed successfully')
-    })
+    fs.unlinkSync(hookFile)
+    spinner.succeed('Gitmoji commit hook removed successfully')
```
carloscuesta, line 15: I'm curious, what did we changed this? 🤔

src/utils/getAbsoluteHooksPath.js
```diff
@@ -0,0 +1,40 @@
...
+      // Use dirname to remove the trailing "/.git" from the path
+      path.dirname(absoluteGitDirCommand.stdout),
+      hooksPath
+    )
+  }
+
+  return path.resolve(absoluteHooksPath, hookName)
+}
+
+export default getAbsoluteHooksPath
```
carloscuesta, line 40: I don't understand why we need to check if it's absolute or not, can we make something like this?

```suggestion
// @flow
import execa from 'execa'
import path from 'path'

const getGitHooksPath = async (hookName: string): string => {
  const gitCoreHooksPath = await execa('git', [
    'config',
    '--get',
    'core.hooksPath'
  ])

  const defaultHooksPath = await execa('git', [
    'rev-parse',
    '--absolute-git-dir'
  ]) + '/hooks'
  
  const hooksPath = gitCoreHooksPath.stdout || defaultHooksPath.stdout

  return path.resolve(hooksPath, hookName)
}

export default getGitHooksPath
```

test/commands/hook.spec.js
```diff
@@ -15,46 +16,62 @@ describe('hook command', () => {
   })
 
   describe('create hook', () => {
-    beforeAll(() => {
-      execa.mockReturnValue({ stdout: stubs.gitAbsoluteDir })
-      hook.create()
+    beforeEach(() => {
+      execa.mockReturnValueOnce({ stdout: stubs.relativeCoreHooksPath })
+      execa.mockReturnValueOnce({ stdout: stubs.gitAbsoluteDir })
```
carloscuesta, line 21: I think that here we're testing the internal implementation of `getAbsoluteHooksPath`, in this spec I would suggest to mock this module and test the implementation on the `getAbsoluteHooksPath` spec

test/setupTests.js
```diff
@@ -5,7 +5,9 @@ jest.mock('path-exists')
 jest.mock('fs')
 jest.mock('ora', () =>
   jest.fn().mockReturnValue({
-    start: jest.fn(),
+    start: jest.fn(function () {
+      return this
+    }),
```
carloscuesta, line 10: 🤔

test/utils/getAbsoluteHooksPath.spec.js
```diff
@@ -0,0 +1,89 @@
+import execa from 'execa'
+// import fs from 'fs'
+import path from 'path'
+
+import getAbsoluteHooksPath from '../../src/utils/getAbsoluteHooksPath'
+import * as stubs from './stubs'
+
+describe('getAbsoluteHooksPath', () => {
```
carloscuesta, line 8: Maybe we can simplify this spec if we refactor the utility!

test/utils/isHookCreated.spec.js
```diff
@@ -1,13 +1,15 @@
...
 import isHookCreated from '../../src/utils/isHookCreated'
 import HOOK from '../../src/commands/hook/hook'
 import * as stubs from './stubs'
 
 describe('isHookCreated', () => {
-  beforeAll(() => {
-    execa.mockReturnValue({ stdout: stubs.gitAbsoluteDir })
+  beforeEach(() => {
+    execa.mockReturnValueOnce({ stdout: stubs.relativeCoreHooksPath })
+    execa.mockReturnValueOnce({ stdout: stubs.gitAbsoluteDir })
```
carloscuesta, line 12: The same as before, we're testing the internal implementation of the util, let's mock the util in this spec!

carloscuesta (commented):
Thanks!

src/utils/getAbsoluteHooksPath.js
flo-sch: Somehow from your previous review, I thought you wanted to always ensure working with an absolute path, but I guess using just a relative path as you illustrate it works too :)

We would get the following results, based on the value of `core.hooksPath`

- `/etc/git/hooks` --> `/etc/git/hooks/prepare-commit-msg`
- `undefined` --> `/path/to/project/.git/hooks/prepare-commit-msg`
- `.git/hooks` --> `.git/hooks/prepare-commit-msg`
- `.husky` --> `.husky/prepare-commit-msg`

When my proposal would instead always return an absolute path:

- `/etc/git/hooks` --> `/etc/git/hooks/prepare-commit-msg`
- `undefined` --> `/path/to/project/.git/hooks/prepare-commit-msg`
- `.git/hooks` --> `/path/to/project/.git/hooks/prepare-commit-msg`
- `.husky` --> `/path/to/project/.husky/prepare-commit-msg`

src/commands/hook/remove/index.js
flo-sch: There were some `UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'fail' of undefined` in the tests, I felt easier to handle that with the `*Sync` API inside an async function

test/setupTests.js
flo-sch: Fix for: `UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'fail' of undefined`

The way spinners are used inside the hooks was by chaining the constructor and start method.
Ora's [start()](https://github.com/sindresorhus/ora#starttext) method returns its instance, allowing to then invoke other methods such as `succeed()` or `fail()`.
But if the mocked implementation of `start()` does not return its instance, then`spinner.fail()` or `spinner.succeed()` fails with the error message above

test/utils/getAbsoluteHooksPath.spec.js
flo-sch: Simpler version of the helper available, I kept the 3 same scenarios though, just simplified a bit the matchers.
Not sure if you want me to rewrite some?

carloscuesta (approved):
Awesome thanks for the work! 🚀

Issue #579, opened 2021-04-22
Gitmoji update fails behind corporate proxy


thtliife:
Hello @carloscuesta!

When trying to use `gitmoji --update` behind a corporate proxy, an ECONNREFUSED error is encountered.

Issue:

node-fetch fails to connect through a proxy in the fetch function call in `utils/getEmojis.js`
Steps to reproduce:
* on a computer behind a proxy, run `gitmoji --update`
* observe error

**OS**:
macOS Big Sur

**gitmoji -v**:
`3.4.0`

**Node version**:
* `14.16.1`
* `15.14.0`
* `16.0.0`

```bash
❯ gitmoji --update
✖ Error: FetchError: request to https://raw.githubusercontent.com/carloscuesta/gitmoji/master/src/data/gitmojis.json failed, reason: connect ECONNREFUSED 185.199.111.133:443
TypeError: Cannot read property 'forEach' of undefined
    at printEmojis (/opt/homebrew/Cellar/gitmoji/3.4.0/libexec/lib/node_modules/gitmoji-cli/lib/utils/printEmojis.js:13:19)
    at /opt/homebrew/Cellar/gitmoji/3.4.0/libexec/lib/node_modules/gitmoji-cli/lib/commands/update/index.js:14:94
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
```

thtliife:
pr incoming with proxy support for node-fetch...

S-Mann:
@thtliife hey! When are you planning to merge this and in which version this will be available? It would be great to pass proxy settings as `params` if needed.

thtliife:
hi @S-Mann 
I am not able to merge pr's in this repo... i am merely a contributor.
The delay on this one, however, was my fault.
I needed to address a couple of issues before merging would be allowed, and write some tests.

PR #580, opened 2021-04-22
Proxy support

thtliife:
## Description

Add proxy agent for node-fetch behind proxy

Issue: #579 

## Tests

- [x] All tests passed.


thtliife:
will redo this with signed commits...

Issue #635, opened 2021-07-06
request to https://gitmoji.dev/api/gitmojis failed


webkong:
Hello @carloscuesta!

git commit error , can not get the gitmojis.json file.

Issue:

OS: macOS 11.4
gitmoji -v 4.1.0
Node: 14.15.5 

**Error Message**:

Error: FetchError: request to https://gitmoji.dev/api/gitmojis failed, reason: connect ECONNREFUSED 76.76.21.21:443
TypeError: Cannot read property 'map' of undefined

PR #636, opened 2021-07-06
Update getEmojis.js

webkong:
## Fix gitmojis url 
Can't get the gitmojis.json file, when used for the first time.

change the url to https://raw.githubusercontent.com/carloscuesta/gitmoji/master/src/data/gitmojis.json

Issue: #635

## Tests

- [x] All tests passed.


carloscuesta:
Hey actually this was changed because users from China couldn't access the raw.github url don't know why it failed to you but not related with the actual endpoint

PR #640, opened 2021-07-14
🔧 Add gitmojis url config

YunYouJun:
## Description

`gitmoji.dev` is not accessible in some regions such as China. See Issue: [network](https://github.com/carloscuesta/gitmoji-cli/issues?q=network).

I think it is necessary to allow custom `GITMOJIS_URL`, so that users can even define their own desired list.

For example, use`https://raw.githubusercontent.com/carloscuesta/gitmoji/master/src/data/gitmojis.json` instead of `https://gitmoji.dev/api/gitmojis`.

## Tests

- [x] All tests passed.


carloscuesta (approved):
Thanks for opening the PR 😊

PR #656, opened 2021-08-03
✨ Introduce option flags for `commit` commands

### LINKED ISSUE #465: [Feature request]: Add --scope option to gitmoji -c command

carloscuesta:
## Description

This PR introduces the possibility of passing options to the `commit` and `hook` commands:

- `--title`
- `--message`
- `--scope`

Example usage:

```
$ gitmoji -c --title="Commit title" --message="Commit message" --scope="SCOPE-123"
```

The feature is intended for "advanced" users who might want to not introduce the prompts on the "interactive" prompt part.

Fixes #465 

## Demo

https://user-images.githubusercontent.com/7629661/128042792-007bd989-04cf-4572-a13c-6c9c77f4c0b2.mov

## Tests

- [x] All tests passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,20 @@ Start the interactive commit client, to auto generate your commit based on your 
 $ gitmoji -c
 ```
 
+##### Options
+
+You can pass default values to the prompts using the following flags:
+
+- `title`: For setting the commit title.
+- `message`: For setting the commit message.
+- `scope`: For setting the commit scope.
+
+Those flags should be used like this:
+
+```bash
+$ gitmoji -c --title="Commit" --message="Message" --scope="Scope"
+```
+
 #### Hook
 
 Run the init option, add your changes and commit them, after that the prompts will begin and your commit message will be built.
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
   },
   "lint-staged": {
     "*.{js}": [
-      "prettier --write src/**/*.js",
+      "prettier --write",
       "git add"
     ]
   },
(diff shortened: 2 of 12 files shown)


carloscuesta (commented):
🚀

Issue #680, opened 2021-09-11
Print failed git command


capsuleman:
Hello @carloscuesta!

Frequently, on my projects, commit fails (sometime I forget to add files, other times pre-commit hooks return error).
It would be nice to have the git commit command print to just copy / paste it the next times. 

Issue:

**OS**:
**gitmoji -v**:
**Node version**:

**Error Message**:

carloscuesta:
Why not using arrow up key instead? It should be faster than copy pasting right?

capsuleman:
When I use `gitmoji -c`, I do not provide any other parameters and enter gitmoji and message in the cli. If I use arrow up key, I can not find again the parameters I filled previously.

PR #681, opened 2021-09-11
Print Git command if it failed

capsuleman:
## Description

If commit fails to execute, cli prints command so it can be copy / paste without executing again gitmoji-cli.
Updated the tests too.

Issue: #680

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/commit/withClient/index.js
+++ b/src/commands/commit/withClient/index.js
@@ -37,8 +37,12 @@ const withClient = async (answers: Answers) => {
   } catch (error) {
     console.error(
       chalk.red(
-        '\nOops! An error ocurred. There is likely additional logging output above.\n'
-      )
+        '\n',
+        'Oops! An error ocurred. There is likely additional logging output above.\n',
+        'You can run the same commit with this command:\n'
+      ),
+      '\t',
+      error.escapedCommand
     )
   }
 }
--- a/test/commands/commit.spec.js
+++ b/test/commands/commit.spec.js
@@ -1,6 +1,7 @@
 import inquirer from 'inquirer'
 import execa from 'execa'
 import fs from 'fs'
+import chalk from 'chalk'
 const mockProcess = require('jest-mock-process')
 
 import configurationVault from '../../src/utils/configurationVault'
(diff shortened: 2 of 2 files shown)


carloscuesta:
Hey! Is there any difference between copying the command from the output of the error and using the arrow up key to replicate the same comand without having to manually copy&pasting it? IMHO using arrow key should be much faster, so I don't really see the point of this 🤔

capsuleman:
> Hey!
> 
> Thanks for opening a PR
> 
> Is there any difference between copying the command from the output of the error and using the arrow up key to replicate the same comand without having to manually copy&pasting it? IMHO using arrow key should be much faster, so I don't really see the point of this thinking

When I use gitmoji -c, I do not provide any other parameters and enter gitmoji and message in the cli. If I use arrow up key, I can not find again the parameters I filled previously.
If there is a quick way without copy/paste, I would be happy to hear it! :-)

carloscuesta:
I think a screenshot/demo about this would be really helpful 😊

capsuleman:
https://user-images.githubusercontent.com/34281913/133271612-c1f9d80e-65c7-481a-8def-c9b0affb3006.mp4

carloscuesta:
Thanks for sharing! But this is the current behaviour right? I was asking for a demo of your change to see what is printed, thank you 🙏🏼

capsuleman:
Oh sorry, I was just wondering if I was using the cli well. With the new version the command would be printed if it fail

https://user-images.githubusercontent.com/34281913/133651823-59186433-54e9-49dc-bac4-b121c00b2dfc.mp4

carloscuesta:
Thanks for sharing the demo! 

Currently, we have a way for specifying the default parameters through the cli. With this you can retry the command faster if it fails as prefills all the questions from the cli.

> Screen Shot 2021-09-19 at 10 06 50

I'm not 100% if this is worth to be added into the cli 😅

capsuleman:
Yes sure, but often I think about the gitmoji, title and message after having typed `gitmoji -c`...
Would you prefer having the command with default values printed? Is there a parameter to provide a gitmoji?

carloscuesta (approved):
Let's merge it 😊

Issue #682, opened 2021-09-11
Adjust commit title length


capsuleman:
Hello @carloscuesta!

Real commit title length should be 50 chars. It should be really calculated according to current gitmoji's length.

Issue:

**OS**:
**gitmoji -v**:
**Node version**:

**Error Message**:

carloscuesta:
Hey! If I'm not mistaken, you're not taking into account the space that is being set between the emoji and the commit title, that's why the value was hardcoded to 48 😅

Also gitmoji.length is not a good way to handle this, because if the user picks the unicode format, the length will be more than 1 but in GitHub the :string: will translate into a single emoji, so I think we can stick with the old approach

Thanks for opening the PR 😊

PR #683, opened 2021-09-11
Compute real title length

capsuleman:
## Description

Set title max length chars to 50, and the calculated according to choosen gitmoji

Issue: #682 

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/commit/prompts.js
+++ b/src/commands/commit/prompts.js
@@ -7,7 +7,7 @@ import getDefaultCommitContent from '../../utils/getDefaultCommitContent'
 import { type CommitOptions } from './index'
 import guard from './guard'
 
-const TITLE_MAX_LENGTH_COUNT: number = 48
+const TITLE_MAX_LENGTH_COUNT: number = 50
 
 inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'))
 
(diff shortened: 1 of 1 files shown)


src/commands/commit/prompts.js
```diff
@@ -56,10 +59,10 @@ export default (gitmojis: Array<Gitmoji>, options: CommitOptions): Array<Object>
       name: 'title',
       message: 'Enter the commit title',
       validate: guard.title,
-      transformer: (input: string) => {
-        return `[${
-          (title || input).length
-        }/${TITLE_MAX_LENGTH_COUNT}]: ${input}`
+      transformer: (input: string, { gitmoji }: { gitmoji: string }) => {
+        return `[${(title || input).length}/${
+          TITLE_MAX_LENGTH_COUNT - gitmoji.length
```
carloscuesta, line 64: Hey! If I'm not mistaken, you're not taking into account the space that is being set between the emoji and the commit title, that's why the value was hardcoded to 48 😅 

Also gitmoji.length is not a good way to handle this, because if the user picks the unicode format, the length will be more than 1 but in GitHub the :string: will translate into a single emoji, so I think we can stick with the old approach

Thanks for opening the PR 😊

PR #777, opened 2022-03-03
Feature/add commit extra args

vinicius73:
## Description

Allow defining extra arguments to be used in every commit.

Issue: #

## Tests

- [x] All tests passed.


vinicius73:
Hi @carloscuesta, can you check it?

carloscuesta:
Hey we already discussed a similar implementation of this into #192 can you explain the use-case of the arguments that you want to pass?

vinicius73:
Hi @carloscuesta, thanks to answer.

Now a days I must use [`--signoff`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s) in all of my commits.

I use gitmoji-cli every time, but I can't add `--signoff` using it.

carloscuesta:
I see, curious why you use the sign-off but not actually signing commits with a verified GPG key. I feel that adding `extraArgs` would make the cli more prone to errors but actually not covering them can make people opt out of using the cli for the same reason 😓 

I think we need to find a solution where we can automatically forward the unknown parameters to the underlying git command so we don't have to write any extra code

Something like this

```
gitmoji -c -- -signoff
```

carloscuesta:
Hey @vinicius73 I'll close this PR as you can setup the way `git` signs off commits in your global `.gitconfig`

Thanks! 🙏🏼

PR #830, opened 2022-05-04
Remove $ shell character from README CLI commands

Christopher-Hayes:
Allow users to copy shell commands using GitHubs code block copy button without needing to remove the shell character on paste.

## Description

Since GitHub has now has a copy button on code bock fields, it makes to avoid prepending shell commands with the `$` shell character. Users have to manually remove the character if they want to copy-paste the commands into their CLI.

This is mainly to make installing easier for users. But, removed it on all code blocks to avoid confusion around inconsistency.

Btw love gitmoji, I use it basically every day on all my projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,19 +20,19 @@ This project provides an easy solution for using [**gitmoji**](https://github.co
 ### npm
 
 ```bash
-$ npm i -g gitmoji-cli
+npm i -g gitmoji-cli
 ```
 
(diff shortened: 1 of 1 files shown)


carloscuesta:
Hey!

I totally agree with you! Thanks for raising a PR and for the great feedback ❤️

carloscuesta (approved):

PR #835, opened 2022-05-13
:sparkles: Adds optional project based emoji file

daschaa:
## Description

The goal of the implemented feature is to enable project specific emoji configuration. The idea is, that gitmoji checks if a `.gitmoji` folder with a `gitmojis.json` (equivalent to the global file) exists in the project directory. If this is the case, the JSON from the project is taken.
The idea behind this is to restrict the emojis to a specific set agreed on by a project team.

## Tests

- [x] All tests passed.


daschaa:
@carloscuesta This pr is now ready for review 😊 Looking forward for your feedback on this! Thanks in advance! ❤️

Issue #893, opened 2022-08-10
gitmoji vs xdg folders


ccoVeille:
Hello @carloscuesta!

I would like gitmoji to respect xdg specifications
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

And so store the cache file in `~/home/.cache/gitmoji/` folder.

carloscuesta:
Hey!

I have no plans on implementing this, I’m quite happy with how `conf` handles storing the configuration of the cli.

If you’re not happy with it you can always fork the project and implement it!

In case you have more feature suggestions, please use the provided issue template, thanks 😊

ccoVeille:
> In case you have more feature suggestions, please use the provided issue template, thanks blush

I'm not sure why the template didn't apply.

realitymolder:
> I'm currently implementing no worries

did you manage to implement that?

ccoVeille:
The author refused everything

https://github.com/carloscuesta/gitmoji-cli/pull/894#issuecomment-1211296857

PR #894, opened 2022-08-10
use .cache/gitmoji folder for cache to respect xdg specifications

ccoVeille:
## Description

Update the script to use a xdg compliant folder for cache file.

Issue: #893

### DIFF
--- a/src/utils/emojisCache.js
+++ b/src/utils/emojisCache.js
@@ -5,7 +5,7 @@ import path from 'path'
 import pathExists from 'path-exists'
 
 export const GITMOJI_CACHE: Object = {
-  FOLDER: '.gitmoji',
+  FOLDER: '.cache/gitmoji',
   FILE: 'gitmojis.json'
 }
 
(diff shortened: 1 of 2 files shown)


carloscuesta:
Hey! Thanks for opening a PR, sorry If I did not explain myself properly, but I don't want to implement this change to the cli,

I'm happy with the current implementation (doesn't depends on another folders such as `.cache`) and I would not like to leave a "useless" `.gitmoji` folder if we change now the path

I'll close this PR, thanks 😊

ccoVeille:
:unamused: but ok.

realitymolder:
> Hey! Thanks for opening a PR, sorry If I did not explain myself properly, but I don't want to implement this change to the cli,
> 
> I'm happy with the current implementation (doesn't depends on another folders such as `.cache`) and I would not like to leave a "useless" `.gitmoji` folder if we change now the path
> 
> I'll close this PR, thanks 😊 

Why not support an env variable to be changeable for the users that DO want to move the files? 
Unfortunately, I had to delete the program until it support the standard...

carloscuesta:
Hey locking this discussion as it was closed feel free to reopen

Remember you can always fork the project if you're not happy with the decisions 😊

Beauty of open source

PR #925, opened 2022-09-27
Fixed vulnerabilities and tests

aaditmshah:
## Description

Most of the tests were failing because we were trying to import ESM `node_modules` in CJS. To fix this, I set the `transformIgnorePatterns` to an empty array and moved the babel config to a separate `babel.config.cjs` file. This ensured that babel transformed all the `node_modules` before running Jest.

The chalk packages uses the `package.json` imports field, which Jest does not understand. The workaround was to use `moduleNameMapper` to map these imports to the correct paths.

The `src/cli.js` file uses `import.meta` which can not be used in CJS files. To fix this issue, I installed the `babel-plugin-transform-import-meta` package and included it in the babel test environment config. This package transforms `import.meta.url` into a CJS equivalent. However, it does not transform `import.meta`. Hence, I manually created a meta object in the source file with the url field set to `import.meta.url`.

I also set the mock implementation of `fs.readFileSync` to the actual implementation of `fs.readFileSync` so that we can read the `package.json` file from `src/cli.js`. In addition, I updated the snapshots as required.

## Tests

- [x] All tests passed.

### DIFF
--- a/test/__snapshots__/cli.spec.js.snap
+++ b/test/__snapshots__/cli.spec.js.snap
@@ -58,6 +58,9 @@ exports[`cli should match meow with cli information 1`] = `
           "type": "boolean",
         },
       },
+      "importMeta": {
+        "url": "file:///home/aaditmshah/Repositories/gitmoji-cli/src/cli.js",
+      },
     },
   ],
 ]
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -1,0 +1,26 @@
+const presets = ['@babel/preset-flow']
+
+const plugins = [
+  [
+    'module-resolver',
+    {
+      alias: {
+        '@utils': './src/utils',
+        '@commands': './src/commands',
+        '@constants': './src/constants'
+      }
+    }
+  ],
+  '@babel/plugin-syntax-import-assertions'
+]
+
+module.exports = {
+  presets,
+  plugins,
+  env: {
+    test: {
+      presets: ['@babel/preset-env', ...presets],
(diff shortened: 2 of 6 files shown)


test/__snapshots__/cli.spec.js.snap
```diff
@@ -58,6 +58,9 @@ exports[`cli should match meow with cli information 1`] = `
           "type": "boolean",
         },
       },
+      "importMeta": {
+        "url": "file:///home/aaditmshah/Repositories/gitmoji-cli/src/cli.js",
```
aaditmshah, line 62: This is an incorrect snapshot.

aaditmshah (commented):
@carloscuesta Updating the snapshots exposed my system file path. Would it be all right to disable snapshot testing for this file?

test/__snapshots__/cli.spec.js.snap
aaditmshah: Fixed it.

aaditmshah:
Hey @carloscuesta, fixed all the issues with ESM testing. This PR is now ready for review.

test/__snapshots__/cli.spec.js.snap
carloscuesta: Yes we should mock this 👀

carloscuesta:
Taking a look into it right now! Thanks for the help 😊

carloscuesta:
Hey!

Made some modifications to the PR but overall looks pretty solid, thanks 🙏🏼

carloscuesta (approved):

PR #926, opened 2022-09-27
♻️ Migrate cli to ESM

carloscuesta:
## Description

This PR updates the `gitmoji-cli` to an ESM so we can update a lot of the dependencies that we're using.

Thanks @aaditmshah for the help! 😊 

## Tests

- [x] All tests passed.

## Pending

- [x] Migrate `getConfiguration` to avoid using `require`

### DIFF
--- a/src/utils/configurationVault/getConfiguration.js
+++ b/src/utils/configurationVault/getConfiguration.js
@@ -1,8 +1,8 @@
 import Conf from 'conf'
 import { cwd } from 'process'
-import pathExists from 'path-exists'
+import { pathExistsSync } from 'path-exists'
 
-import { CONFIG, EMOJI_COMMIT_FORMATS } from '@constants/configuration'
+import { CONFIG, EMOJI_COMMIT_FORMATS } from '@constants/configuration.js'
 
 const DEFAULT_CONFIGURATION = {
   [CONFIG.AUTO_ADD]: false,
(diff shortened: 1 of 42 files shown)


src/utils/configurationVault/getConfiguration.js
```diff
@@ -30,11 +30,11 @@ const getConfiguration = (): { get: Function, set: Function } => {
     const packageJson = `${cwd()}/package.json`
     const configurationFile = `${cwd()}/.gitmojirc.json`
 
-    if (pathExists.sync(packageJson) && require(packageJson)?.gitmoji) {
+    if (pathExistsSync(packageJson) && require(packageJson)?.gitmoji) {
       return require(packageJson).gitmoji
```
carloscuesta, line 34: We need to migrate this to `readFileSync` as `require` can't be used

src/utils/configurationVault/getConfiguration.js
```diff
@@ -30,11 +30,11 @@ const getConfiguration = (): { get: Function, set: Function } => {
...
     }
 
-    if (pathExists.sync(configurationFile) && require(configurationFile)) {
+    if (pathExistsSync(configurationFile) && require(configurationFile)) {
```
carloscuesta, line 37: We need to migrate this to `readFileSync` as `require` can't be used

src/utils/configurationVault/getConfiguration.js
aaditmshah: Completed this task in #928.

src/utils/configurationVault/getConfiguration.js
aaditmshah: Completed this task in #928.

PR #1000, opened 2023-02-01
🚸 Improve error message when doing commits

ekwoka:
This PR adjusts the command sent for `gitmoji -c` when the `automatically add` is enabled.

Before: if the command failed, it would return a retry command without the `a` preventing the very simply copy+paste to resume after a failure.

Now it will return that in the error when that setting is enabled.

### DIFF
--- a/src/commands/commit/withClient/index.js
+++ b/src/commands/commit/withClient/index.js
@@ -22,18 +22,13 @@ const withClient = async (answers: Answers): Promise<void> => {
         )
       )
     }
+    const autoAdd = configurationVault.getAutoAdd()
 
-    if (configurationVault.getAutoAdd()) await execa('git', ['add', '.'])
+    if (autoAdd) await execa('git', ['add', '.'])
 
     await execa(
       'git',
-      [
-        'commit',
-        configurationVault.getAutoAdd() ? '-am' : '-m',
-        title,
-        '-m',
-        answers.message
-      ],
+      ['commit', autoAdd ? '-am' : '-m', title, '-m', answers.message],
       {
         buffer: false,
         stdio: 'inherit'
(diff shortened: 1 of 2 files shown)


src/commands/commit/withClient/index.js
```diff
@@ -25,10 +25,20 @@ const withClient = async (answers: Answers): Promise<void> => {
...
 
-    await execa('git', ['commit', '-m', title, '-m', answers.message], {
-      buffer: false,
-      stdio: 'inherit'
-    })
+    await execa(
+      'git',
+      [
+        'commit',
+        configurationVault.getAutoAdd() ? '-am' : '-m',
```
carloscuesta, line 32: `-am` here technically is not needed since we're running `git add .` if the option is enabled but I guess it's a way of improving the error handling, can we extract this into a variable?

```suggestion
        configurationVault.getAutoAdd() ? '-am' : '-m',
```

carloscuesta: https://github.com/carloscuesta/gitmoji-cli/blob/5651277b98e949b64d4dc643afe610cfa1cbc929/src/commands/commit/withClient/index.js#L14

```js
const isAutoAddEnabled = configurationVault.getAutoAdd()
```

carloscuesta (commented):
Hey!

Thanks for raising the PR

src/commands/commit/withClient/index.js
ekwoka: Definitely. I did see how the way it works is addequate to solving the goals of the command, just that the error handling ends up being not QUITE right.

Just changing the command like these seemed the best way to just tweak the functionality to get to the goal.

carloscuesta (approved):
LGTM! 🚀

Issue #1038, opened 2023-03-26
within hooks use command  `npx gitmoji` instead of single `gitmoji`


AndreasAugustin:
### Discussion

Hello,

sometimes it is useful not installing a npm tool globally.
Instead e.g. install it locally in a repo.

To be able to e.g. execute binaries you can use [npx](https://www.npmjs.com/package/npx) which is shipped per default with `npm` (i think since npm v7).

With the current version of `gitmoji-cli`  the git hook looks like

```bash
'#!/bin/sh\n# gitmoji as a commit hook\n' +
    'exec < /dev/tty\ngitmoji --hook $1 $2\n'
```

which needs to have `gitmoji-cli` globally installed.

A change to

```bash
'#!/bin/sh\n# gitmoji as a commit hook\n' +
    'exec < /dev/tty\nnpx gitmoji --hook $1 $2\n'
```

would use `npx` to call `gitmoji`.

Related to [docs](https://www.npmjs.com/package/npx)

_By default, npx will check whether <command> exists in $PATH, or in the local project binaries, and execute that._

This would be a nice way to make the solution more portable for multi user projects because the dependency to `gitmoji-cli` can be pinned with a special version in the `package.json`.

**Possible issues:**

- user does not have `npx` installed
  - `npx` comes with `npm` per default, I think starting with v7
  - not sure about the supported brew install.
  - not sure how it is working for yarn users

**Advantages**

- no need to  install `gitmoji-cli` global on the system which helps to keep the system clean
- global installation still supported (if npx is installed)
- portable -> will be installed on each dev system locally because added as dev dependency in the right version because it is part of `package.json`
  - personally I am part of many different projects using many different versions of tools. Often it is needed that the members have the right versions installed.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/carloscuesta/gitmoji-cli/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/carloscuesta/gitmoji-cli/blob/master/.github/CONTRIBUTING.md).
- [X] Check that there isn't already an issue requesting the same feature.

carloscuesta:
Hey!

Thanks for opening the issue, I'm aware of `npx` and its benefits. I think it's definitely something we can explore:

> not sure about the supported brew install.

The `gitmoji-cli` package depends on `node` as you can see here on the formulae page: https://formulae.brew.sh/formula/gitmoji

[...]

PR #1039, opened 2023-03-26
Feat/use npx

AndreasAugustin:
## Description

This will add `npx`  to the hook `gitmoji` command.
Discussion started in #1038 .

Issue: #1038 

## Tests

Related test is updated
- [x] All tests passed.

### DIFF
--- a/src/commands/hook/hook.js
+++ b/src/commands/hook/hook.js
@@ -4,7 +4,7 @@ const HOOK: Object = {
   FILENAME: 'prepare-commit-msg',
   CONTENTS:
     '#!/bin/sh\n# gitmoji as a commit hook\n' +
-    'exec < /dev/tty\ngitmoji --hook $1 $2\n'
+    'exec < /dev/tty\nnpx gitmoji --hook $1 $2\n'
 }
 
 export default HOOK
--- a/test/commands/__snapshots__/hook.spec.js.snap
+++ b/test/commands/__snapshots__/hook.spec.js.snap
@@ -5,7 +5,7 @@ exports[`hook command should match hook configuration 1`] = `
   "CONTENTS": "#!/bin/sh
 # gitmoji as a commit hook
 exec < /dev/tty
-gitmoji --hook $1 $2
+npx gitmoji --hook $1 $2
 ",
   "FILENAME": "prepare-commit-msg",
   "PERMISSIONS": 509,


src/commands/hook/hook.js
```diff
@@ -4,7 +4,7 @@ const HOOK: Object = {
   FILENAME: 'prepare-commit-msg',
   CONTENTS:
     '#!/bin/sh\n# gitmoji as a commit hook\n' +
-    'exec < /dev/tty\ngitmoji --hook $1 $2\n'
+    'exec < /dev/tty\nnpx gitmoji --hook $1 $2\n'
```
carloscuesta, line 7: Hey!

Let's change this to:

```suggestion
    'exec < /dev/tty\nnpx -c "gitmoji" --hook $1 $2\n'
```

The package name published on the registry is `gitmoji-cli` therefore if we use `npx gitmoji` the command will suggest installing `gitmoji` which is a different package 😅 

Making this change will allow the hook resolving the `gitmoji` cli command either as a local dependency or a global one

carloscuesta (commented):
Thanks!

src/commands/hook/hook.js
AndreasAugustin: > ah right.. if it is not installed locally or globally it won't work. But I think it should be
> 
> ```shell
> npx --package=gitmoji-cli gitmoji
> ```

further update for the script:

```sh
#!/usr/bin/env sh                                                                                                                                                                                             

if npx -v >/dev/null
then  npx --package=gitmoji-cli gitmoji --hook $1 $2
else  gitmoji --hook $1 $2
fi

```

- small change in shebang to make it more portable
- check in the hook script
  - is `npx` installed -> use `npx` 
  - not installed -> use global installed version

AndreasAugustin: @carloscuesta updated the code.
Now using `npx -c gitmoji` if npx is installed. Else fallback to `gitmoji`.
But another thought. Think it makes sense to change the check to `if npx -c gitmoji -v >/dev/null `. Then it checks in addiition if `gitmoji` can be called with `npx -c`.
Currently am not sure if there is an issue if e.g. `npx` is installed and `gitmoji` is installed with e.g. `brew`.

carloscuesta (approved):
LGTM! 🚀 

Thanks 🙏🏼

src/commands/hook/hook.js
```diff
@@ -3,8 +3,13 @@ const HOOK: Object = {
   PERMISSIONS: 0o775,
   FILENAME: 'prepare-commit-msg',
   CONTENTS:
-    '#!/bin/sh\n# gitmoji as a commit hook\n' +
-    'exec < /dev/tty\ngitmoji --hook $1 $2\n'
+    '#!/usr/bin/env sh\n# gitmoji as a commit hook\n' +
+    'if npx -v >/dev/null\n' +
```
carloscuesta, line 7: Wait! I think we need to redirect std error as well, otherwise if the command is not available it will show an error:

I believe we should check the status code of the command:

https://www.cyberciti.biz/faq/bash-get-exit-code-of-command/

```suggestion
    'if npx -v >&/dev/null\n' +
```

AndreasAugustin: @carloscuesta thanks for the trigger. Sorry I totally forgot about that PR..
Updated your suggestions.

AndreasAugustin:
@carloscuesta am not able to run the `ci` actions. An approval of a maintainer is awaited.
Furthermore am not sure about the workflow.

If you are fine. Am happy to merge, or also am happy if you merge :+1:

Issue #1040, opened 2023-03-26
support `git commit`  options within `gitmoji -c`


AndreasAugustin:
### Description of the problem

Currently it is not possible to use `git commit` flags and options when calling `gitmoji -c`.

Because of that some super useful options like `git commit -S`  to sign the commits is not possible to be used within that command.

### Solution

add the needed flags as optional parameters to the already supported flags.

### Alternatives

Currently  it is possible to add some of the flags to the git global settings. E.g. with the example for `-S`

```bash
git config --global commit.gpgsign true
```

This is possible but somehow not nice

### Additional context

For me in special the `-S` option is needed and useful. In many projects I  want and need to sign commits and should be considered as best practice. For more information please consider https://dev.to/andreasaugustin/git-how-and-why-to-sign-commits-35dn

There are many more useful options which are currently suppressed.  Happy to discuss.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/carloscuesta/gitmoji-cli/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/carloscuesta/gitmoji-cli/blob/master/.github/CONTRIBUTING.md).
- [X] Check that there isn't already an issue requesting the same feature.

carloscuesta:
Hey!

Thanks for opening an issue, completely agree, we should support all the same options that `git` accepts as a parameter without having to parse them on the cli. Any unknown option should be forwarded to the `git` binary.

Perhaps we can fix that by using `allowUnknownFlags` and passing those to `git`

https://github.com/sindresorhus/meow

AndreasAugustin:
> Hey!
> 
> Thanks for opening an issue, completely agree, we should support all the same options that `git` accepts as a parameter without having to parse them on the cli. Any unknown option should be forwarded to the `git` binary.
> 
> Perhaps we can fix that by using `allowUnknownFlags` and passing those to `git`
> 
> https://github.com/sindresorhus/meow
> 
> Do you want to raise a pull-request?

Yes, I can raise a PR. Just cannot promise an ETA. Think I will be at least quite busy the next days (but it shouldn't be a huge change)

Issue #1037, opened 2023-03-24
Disable hook if commit message already starts with emoji


thdk:
### Description of the problem

Context: automating bumping versions of node packages

```sh
❯ gitmoji -i
❯ git commit -a -m "🔖 Bump version to $npm_package_version"
? Choose a gitmoji: 🔖  - Release / Version tags.
? Enter the commit title [18/48]: 🔖 Bump version to 0.1.0
? Enter the commit message: 
[feature/abc 468583a] 🔖 🔖 Bump version to 0.1.0
 2 files changed, 4 insertions(+), 4 deletions(-)
```

Could we detect in the prepare commit message hook that the commit message already contains an emoji and if so skip the emoji prompt?

Or still show the prompt but preselect the result with the emoji from the commit message?

### Solution

```
const commit = (options: CommitOptions): Function => {
  if (options.mode === COMMIT_MODES.HOOK) {
  registerHookInterruptionHandler()
    if (options.message?.startsWith(anyEmoji)) {
        commit(options)
     else {
        
        return cancelIfNeeded().then(() => promptAndCommit(options))
    }
  }

  return promptAndCommit(options)
}
```

### Alternatives

`git commit --no-verify` doesn't seem to prevent the gitmoji hook.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/carloscuesta/gitmoji-cli/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/carloscuesta/gitmoji-cli/blob/master/.github/CONTRIBUTING.md).
- [X] Check that there isn't already an issue requesting the same feature.

carloscuesta:
Hey!

Thanks for opening the issue, I think it makes sense but we should probably aim to support the `--no-verify` option instead of doing this "magic" of skipping the hook if the emoji is provided already 

Feel free to raise a Pull Request! 🙏🏼

carloscuesta:
Hey! So I took at stab at this and found out it's not possible to use `--no-verify` for the `prepare-commit-msg` hook (it is ignored) [as mentioned in the docs](https://git-scm.com/docs/githooks#_prepare_commit_msg)

<img width="681" alt="CleanShot 2023-05-31 at 17 44 37@2x" src="https://github.com/carloscuesta/gitmoji-cli/assets/7629661/59a8078f-5772-4df9-b16c-d927e015abae">

I implemented the PR as an alternative solution to this.

[...]

PR #1045, opened 2023-03-28
Feat/support git commit options

AndreasAugustin:
## Description

PR is for discussion. Tests need to be repaired.
no suppressing of additional git tags anymore

Issue: #1040 

## Tests

- [x] All tests passed.


test/commands/commit.spec.js
```diff
@@ -80,7 +80,7 @@ describe('commit command', () => {
         expect(execa).toHaveBeenCalledWith('git', ['add', '.'])
       })
 
-      it('should call execa with the commit command based on answers', () => {
+      it.skip('should call execa with the commit command based on answers', () => {
```
AndreasAugustin, line 83: :warning:  need to be repaired and enabled again

src/utils/findGitmojiCommand.js
```diff
@@ -1,16 +1,26 @@
...
 
   if (commandsWithOptions.includes(command)) {
+    const gitFlags = omit(
+      flags,
+      'm',
+      'add',
+      ...Object.values(COMMIT_MESSAGE_PARTS),
+      ...Object.values(FLAGS)
+    )
+    // TODO(anau) check if message, mode, scope, title in git flags (-m ":mode: (scope) title" -m "message" -> 'm' is array if 2 times) includes #1037
```
AndreasAugustin, line 16: here the check for #1037 is possible to be included.
Additional  it makes sense to add a check if `gitmoji` is called e.g. with
`gitmoji -c -m ":tada: (bootstrap) bootstrapping project"` (maybe in that case it does not make sense to use gitmoji, but possible with hook, or automation)

src/utils/findGitmojiCommand.js
```diff
@@ -19,14 +29,23 @@ const getOptionsForCommand = (command: ?string, flags: Object): ?Object => {
...
     .find((flag) => options[flag])
-  const commandOptions = getOptionsForCommand(commandFlag, flags)
 
+  const commandOptions = getOptionsForCommand(commandFlag, flags)
   return options[commandFlag]
     ? options[commandFlag](commandOptions)
     : cli.showHelp()
 }
 
+const omit = (obj: Object, ...props: mixed[]) => {
```
carloscuesta, line 43: Hey! Not sure if I follow, can we use here?

https://github.com/sindresorhus/meow#allowunknownflags

To forward all of the unknown flags to the `git` command?

AndreasAugustin: @carloscuesta have you found some time to do some further investigations? Am still sure that https://github.com/sindresorhus/meow#allowunknownflags does not help with the current code because currently the unknown flags are already forwarded but the business logic within the code has a mapping part and does (not yet) map unknown flags

PR #1072, opened 2023-05-05
✨ Allow user to disable message prompt

segersniels:
## Description

This PR updates the configuration vault, adding new functions to manage message and scope prompts. Most of the time you don't need a commit body and only need the title for quick commits. So to improve UX allow the user to disable that option so it only prompts for a gitmoji and title.

## Tests

- [x] All tests passed.

### DIFF
--- a/src/utils/configurationVault/getConfiguration.js
+++ b/src/utils/configurationVault/getConfiguration.js
@@ -16,13 +16,27 @@ const DEFAULT_CONFIGURATION = {
 const LOCAL_CONFIGURATION: typeof Conf = new Conf({
   projectName: 'gitmoji',
   schema: {
-    [CONFIG.AUTO_ADD]: { type: 'boolean' },
-    [CONFIG.EMOJI_FORMAT]: {
-      enum: Object.values(EMOJI_COMMIT_FORMATS)
+    [CONFIG.AUTO_ADD]: {
+      type: 'boolean',
+      default: DEFAULT_CONFIGURATION[CONFIG.AUTO_ADD]
     },
-    [CONFIG.SCOPE_PROMPT]: { type: 'boolean' },
-    [CONFIG.MESSAGE_PROMPT]: { type: 'boolean' },
-    [CONFIG.GITMOJIS_URL]: { type: 'string', format: 'url' }
+    [CONFIG.EMOJI_FORMAT]: {
+      enum: Object.values(EMOJI_COMMIT_FORMATS),
+      default: DEFAULT_CONFIGURATION[CONFIG.EMOJI_FORMAT]
+    },
+    [CONFIG.SCOPE_PROMPT]: {
+      type: 'boolean',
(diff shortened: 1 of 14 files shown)


segersniels:
Not sure if this needs a test.

segersniels:
> Hey!
> 
> You introduced a configuration option, but it's not doing anything. Now you'll need to skip the prompt when the option is enabled, like this:
> 
> https://github.com/carloscuesta/gitmoji-cli/blob/20aebc39a00d8745da60a0cc6143d9c781fea483/src/commands/commit/prompts.js#L49-L57

Haha, exactly 🙈 Was pumping out these PRs that I probably missed a thing or two. Good catch.

src/utils/configurationVault/getConfiguration.js
```diff
@@ -9,6 +9,7 @@ const DEFAULT_CONFIGURATION = {
   [CONFIG.AUTO_ADD]: false,
   [CONFIG.EMOJI_FORMAT]: EMOJI_COMMIT_FORMATS.CODE,
   [CONFIG.SCOPE_PROMPT]: false,
+  [CONFIG.MESSAGE_PROMPT]: true,
```
carloscuesta, line 12: If someone already has a configuration initialised (locally) they'll have to run the `gitmoji -g` again to set this new key.

By default after adding this the message prompt will disappear for everyone 😓 

I'm wondering if we should somehow take this into account here:

https://github.com/carloscuesta/gitmoji-cli/blob/276713b7f6a30b7fd3722aa481543644099d5b6d/src/utils/configurationVault/getConfiguration.js#L55-L62

So we can provide the fallback for values that are newly introduced

segersniels: Easiest solution would be to utilise the `default` schema property from `conf` but that won't apply to users that use a local config in their projects. I guess we could do both, provide the defaults in schema from the `DEFAULT_CONFIG` as well as falling back to `DEFAULT_CONFIG[key]` in case `configuration[key]` doesn't return anything.

segersniels: Check out e06760deb4a0cf2911b515bf4503456d622d1aeb

carloscuesta: Yep! That makes a lot of sense 🙏🏼 🚀

src/utils/configurationVault/getConfiguration.js
```diff
@@ -59,7 +75,7 @@ const getConfiguration = (): { get: Function, set: Function } => {
           ? resolvedConfiguration
           : DEFAULT_CONFIGURATION
 
-      return configuration[key]
+      return configuration[key] ?? DEFAULT_CONFIGURATION[key]
```
carloscuesta, line 78: Should we add a small check in the spec for this? 👀

carloscuesta (approved):
LGTM! 🚀 

Thanks ❤️

Issue #1070, opened 2023-05-05
Expose flags as actual subcommands


segersniels:
### Description of the problem

Not sure if I'm only speaking for myself but I find the way that `gitmoji` is set up a bit conflicting with how the usual CLI tool is made. 

### Solution

Adding actual subcommands `gitmoji commit` to behave exactly the same as their flags seems like a logical move in the right direction for better UX.

1. `gitmoji commit`
2. `gitmoji update`
3. `gitmoji config`
4. `gitmoji list`
5. ...

### Alternatives

https://github.com/segersniels/gitmoji

### Validations

- [X] Follow our [Code of Conduct](https://github.com/carloscuesta/gitmoji-cli/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/carloscuesta/gitmoji-cli/blob/master/.github/CONTRIBUTING.md).
- [X] Check that there isn't already an issue requesting the same feature.

PR #1073, opened 2023-05-05
✨ Expose flags as subcommands

segersniels:
## Description

This PR introduces the usage of subcommands alongside the usual flags. This means now eg. `gitmoji commit` is now usable as well as `gitmoji -c`.

Issue: #1070 

## Tests

- [x] All tests passed.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -17,7 +17,7 @@ updateNotifier({ pkg: JSON.parse(packageJson) }).notify({ isGlobal: true })
 const cli = meow(
   `
   Usage
-    $ gitmoji
+    $ gitmoji [option] [command]
   Options
     --${FLAGS.COMMIT}, -c    Interactively commit using the prompts
     --${FLAGS.CONFIG}, -g    Setup gitmoji-cli preferences.
@@ -27,6 +27,14 @@ const cli = meow(
     --${FLAGS.SEARCH}, -s    Search gitmojis
     --${FLAGS.UPDATE}, -u    Sync emoji list with the repo
     --${FLAGS.VERSION}, -v   Print gitmoji-cli installed version
+  Commands
+    commit          Interactively commit using the prompts
+    config          Setup gitmoji-cli preferences.
+    init            Initialize gitmoji as a commit hook
+    list            List all the available gitmojis
+    remove          Remove a previously initialized commit hook
+    search          Search gitmojis
+    update          Sync emoji list with the repo
   Examples
     $ gitmoji -l
     $ gitmoji bug linter -s
--- a/src/utils/findGitmojiCommand.js
+++ b/src/utils/findGitmojiCommand.js
@@ -2,6 +2,10 @@
 import COMMIT_MODES from '@constants/commit'
 import FLAGS from '@constants/flags'
 
+const isSupportedCommand = (command: ?string, options: Object): boolean => {
+  return Object.keys(options).includes(command)
+}
+
 const getOptionsForCommand = (command: ?string, flags: Object): ?Object => {
   const commandsWithOptions = [FLAGS.COMMIT, FLAGS.HOOK]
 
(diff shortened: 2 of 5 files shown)


segersniels:
Not sure if I missed something important but seems like this should be sufficient to support both commands as flags and subcommands

src/cli.js
```diff
@@ -17,7 +17,7 @@ updateNotifier({ pkg: JSON.parse(packageJson) }).notify({ isGlobal: true })
 const cli = meow(
   `
   Usage
-    $ gitmoji
+    $ gitmoji [options] [command]
   Options
```
carloscuesta, line 21: I believe at some point we might want deprecate the options 👀

src/utils/findGitmojiCommand.js
```diff
@@ -19,14 +23,25 @@ const getOptionsForCommand = (command: ?string, flags: Object): ?Object => {
...
 
-  return options[commandFlag]
-    ? options[commandFlag](commandOptions)
-    : cli.showHelp()
+  if (!command && cli.input.length) {
+    // See if the user provided the command as a subcommand
+    command = cli.input[0]
+  }
+
+  // Unable to determine requested command so display help
```
carloscuesta, line 37:
```suggestion
```

src/utils/findGitmojiCommand.js
```diff
@@ -19,14 +23,25 @@ const getOptionsForCommand = (command: ?string, flags: Object): ?Object => {
 
 const findGitmojiCommand = (cli: any, options: Object): void => {
   const flags = cli.flags
-  const commandFlag = Object.keys(flags)
+
+  // Attempt to get the command out of the flags
```
carloscuesta, line 27:
```suggestion
```

src/utils/findGitmojiCommand.js
```diff
@@ -19,14 +23,25 @@ const getOptionsForCommand = (command: ?string, flags: Object): ?Object => {
...
+  let command = Object.keys(flags)
     .map((flag) => flags[flag] && flag)
     .find((flag) => options[flag])
```
carloscuesta, line 30: Maybe we can do:

```suggestion
    const command = Object.keys(flags)
    .map((flag) => flags[flag] && flag)
    .find((flag) => options[flag]) || cli.input[0]
```

src/utils/findGitmojiCommand.js
```diff
@@ -19,14 +23,25 @@ const getOptionsForCommand = (command: ?string, flags: Object): ?Object => {
...
-  const commandOptions = getOptionsForCommand(commandFlag, flags)
 
-  return options[commandFlag]
-    ? options[commandFlag](commandOptions)
-    : cli.showHelp()
+  if (!command && cli.input.length) {
+    // See if the user provided the command as a subcommand
+    command = cli.input[0]
+  }
+
```
carloscuesta, line 36:
```suggestion
```

carloscuesta (commented):
Good one!

Let's update the spec file here:

https://github.com/carloscuesta/gitmoji-cli/blob/20aebc39a00d8745da60a0cc6143d9c781fea483/test/utils/findGitmojiCommand.spec.js#L6

Thanks! ❤️

src/cli.js
segersniels: Could happen in a different PR, let's first guarantee everything works as intended.

carloscuesta: Yeah sure, also we would need to release it as a major

test/__snapshots__/cli.spec.js.snap
```diff
@@ -5,7 +5,7 @@ exports[`cli should match meow with cli information 1`] = `
   [
     "
   Usage
-    $ gitmoji
+    $ gitmoji [options] [command]
```
carloscuesta, line 8:
```suggestion
    $ gitmoji [option] [command]
```

test/utils/findGitmojiCommand.spec.js
```diff
@@ -6,20 +6,38 @@ import * as stubs from './stubs'
 describe('findGitmojiCommand', () => {
   stubs.commands.map((command) => {
```
carloscuesta, line 7: I believe we can switch to [test.each](https://jestjs.io/docs/api#testeachtablename-fn-timeout) to make this simpler ✨

test/utils/findGitmojiCommand.spec.js
```diff
@@ -6,20 +6,38 @@ import * as stubs from './stubs'
...
+          })
         })
       })
     }
 
-    it(`it should match ${command} command`, () => {
-      findGitmojiCommand(stubs.cliMock({ [command]: true }), stubs.optionsMock)
-      expect(stubs.optionsMock[command]).toHaveBeenCalled()
+    describe('with options', () => {
+      it(`it should match ${command} command`, () => {
```
carloscuesta, line 27:
```suggestion
      it(`it should match ${command} option`, () => {
```

carloscuesta (commented):
Awesome!

A small detail, thanks ❤️

carloscuesta (approved):
Awesome! Let's ship this 🚀 

Thanks ❤️

PR #1075, opened 2023-05-07
🐛 Filter out `search` from query when using subcommands

segersniels:
## Description

With the introduction of the subcommand when you use `gitmoji search bug linter` it will always include the `search` keyword in the query which is a regression. This PR ensures the `search` input is filtered out of `cli.input` when the command is executed through a subcommand.

It changes how the `search` command is handled. Instead of going over each input from within `cli.js` we now determine it in `findGitmojiCommand` and pass the correct option to the command depending on the situation. Where it is then further handled by the `search` command itself.

## Tests

- [x] All tests passed.

### DIFF
--- a/src/commands/search/index.js
+++ b/src/commands/search/index.js
@@ -3,10 +3,21 @@ import filterGitmojis from '@utils/filterGitmojis'
 import getEmojis from '@utils/getEmojis'
 import printEmojis from '@utils/printEmojis'
 
-const search = (query: string): Promise<void> => {
+type SearchOptions = {
+  query?: string[]
+}
+
+const search = (options: SearchOptions): Promise<void> => {
   return getEmojis()
-    .then((gitmojis) => filterGitmojis(query, gitmojis))
-    .then((gitmojisFiltered) => printEmojis(gitmojisFiltered))
+    .then((gitmojis) => {
+      if (!options.query) {
+        return printEmojis(gitmojis)
+      }
+
+      for (const query of options.query) {
+        printEmojis(filterGitmojis(query, gitmojis))
+      }
+    })
     .catch((err) => console.error(err))
 }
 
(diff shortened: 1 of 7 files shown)


segersniels:
@carloscuesta probably good to get this one merged at one point as well since there's a regression when searching through the subcommand

src/utils/findGitmojiCommand.js
```diff
@@ -6,34 +6,67 @@ const isSupportedCommand = (command: ?string, options: Object): boolean => {
...
+
+  if (!command) {
+    type = 'command'
+    command = input[0]
+  }
+
+  return {
+    type,
+    command
+  }
```
carloscuesta, line 30: Let's refactor this into:

```suggestion
  const command = Object.keys(flags)
    .map((flag) => flags[flag] && flag)
    .find((flag) => options[flag])

  return command
    ? {
      type: 'flag',
      command
    }
    : {
      type: 'command',
      command: input[0]
    }
```

carloscuesta (commented):
Sorry for the delay!

Thanks for the PR 🙏🏼

src/commands/search/index.js
```diff
@@ -3,10 +3,21 @@ import filterGitmojis from '@utils/filterGitmojis'
...
+  query?: string[]
+}
+
+const search = (options: SearchOptions): Promise<void> => {
   return getEmojis()
-    .then((gitmojis) => filterGitmojis(query, gitmojis))
-    .then((gitmojisFiltered) => printEmojis(gitmojisFiltered))
+    .then((gitmojis) => {
+      if (!options.query) {
+        return printEmojis(gitmojis)
```
carloscuesta, line 14: Hey! Can we add a test case for this?

segersniels: Oh sorry, I completely forgot this PR and comment. Sorry for the delay!

PR #1076, opened 2023-05-07
📝 Mention correct default value for message prompt

segersniels:
## Description

With the introduction of the message prompt the new default values were added to the `README` but with incorrect values. Not sure if intentional but might be better to reflect the actual default values.

## Tests

- [x] All tests passed.


carloscuesta:
Hey! I think it's fine, it's just an "example" that you can set a boolean value as that preferences, we're not referencing the defaults 🙏🏼

Issue #1108, opened 2023-05-30
Performance improvements


carloscuesta:
### Describe the bug

Hey! 👋🏼 

Our cli can be improved when running and executing commands. In some scenarios and depending on the operating system some commands can take up to 1s to give some output.

See: https://github.com/carloscuesta/gitmoji-cli/issues/1096

### Reproduction

This can be reproduced using the [time](https://man7.org/linux/man-pages/man1/time.1.html) command:

```
time gitmoji --help
time gitmoji --list
time gitmoji --search bug
```

As a result you'll get an output with the seconds each operation take to complete (for the given commands you'll get something like:

```
gitmoji --help  0.41s user 0.08s system 115% cpu 0.425 total
gitmoji --list  0.46s user 0.10s system 62% cpu 0.900 total
gitmoji --search bug  0.45s user 0.08s system 119% cpu 0.448 total
```

### System Info

```Shell
N/A
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/carloscuesta/gitmoji-cli/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/carloscuesta/gitmoji-cli/blob/master/.github/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating duplicates.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

segersniels:
Using dynamic imports might result in some easy wins for the initial startup.

```
gitmoji --help  0.41s user 0.06s system 108% cpu 0.426 total
```

```diff
diff --git a/src/cli.js b/src/cli.js
index 4982d27..4aba009 100755
--- a/src/cli.js
+++ b/src/cli.js
@@ -6,7 +6,6 @@ import { readFileSync } from 'fs'
 
 import FLAGS from '@constants/flags'
 import findGitmojiCommand from '@utils/findGitmojiCommand'
-import commands from './commands'
 
 const packageJson: Object = readFileSync(
   new URL('../package.json', import.meta.url)
@@ -56,14 +55,18 @@ const cli = meow(
 )
 
 export const options = ({
-  [FLAGS.COMMIT]: (options: Object) => commands.commit(options),
-  [FLAGS.CONFIG]: () => commands.config(),
[...]

PR #1126, opened 2023-06-16
:zap: Use dynamic imports instead of static imports

segersniels:
### Description
This PR updates the imports from static imports to dynamic imports which should result in some easy speed improvements.

### Changed
- Improved startup speed by utilising dynamic imports for each command only when they are needed. This improves the initial startup time because it doesn't have to load in all the other commands that are unused.

### Issues
- https://github.com/carloscuesta/gitmoji-cli/issues/1108

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -5,7 +5,6 @@ import { readFileSync } from 'fs'
 
 import FLAGS from '@constants/flags'
 import findGitmojiCommand from '@utils/findGitmojiCommand'
-import commands from '@commands/index'
 import type { CommitOptions } from '@commands/commit'
 import type { SearchOptions } from '@commands/search'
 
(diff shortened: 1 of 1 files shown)


segersniels:
```
gitmoji  0.42s user 0.07s system 112% cpu 0.431 total
node lib/cli.js  0.16s user 0.03s system 78% cpu 0.253 total
```

```
gitmoji search bug linter  0.45s user 0.11s system 80% cpu 0.695 total
node lib/cli.js search big linter  0.32s user 0.05s system 85% cpu 0.430 total
```

```
gitmoji list  0.44s user 0.07s system 88% cpu 0.580 total
node lib/cli.js list  0.31s user 0.05s system 92% cpu 0.383 total
```

src/cli.ts
```diff
@@ -59,14 +58,20 @@ const cli = meow(
 )
 
 export const options = {
```
carloscuesta, line 60: This is great! It improves startup time, but I'm wondering how much it affects the overall execution time of the command? So we can understand the trade-offs of the solution 👀

segersniels: It shouldn't affect anything. It defers loading to when it's needed instead of pulling everything in memory from the get go. Code execution following the dynamic import proceeds like usual. It just gets to executing a lot faster since it doesn't have to load in all the other commands and their dependencies.

Only downside of dynamic importing AFAIK is the fact that tree shaking won't be available for the commands and their dependencies.

carloscuesta (approved):

Issue #1247, opened 2024-01-15
Review default emoji format


mikelorant:
### Discussion

There are significant inconsistencies with some Git remote repository providers such as BitBucket when displaying Emoji shortcodes. BitBucket does display shortcodes for the commit title, but not the PR title and list of PRs.

There is also requirement to modify the pager (or other tricks) to allow `git log` to display correctly.

After investigation it became clear for the reasoning why shortcodes were chosen as the default.

> The reason behind this was to be able to use the shortcode to favour the readability on certain terminals that do not support Emojis!
https://github.com/carloscuesta/gitmoji-cli/issues/630#issuecomment-869215322

While the decision may have been appropriate 8 year ago when `gitmoji-cli` was first released, it does beg the question, is it worth considering changing the default behaviour now?

Terminals are considerably more advanced on all platforms and the standard Gitmoji set is simple enough that all terminals would display them correctly nowadays. I am at a loss to come up with any other reason to keep the shortcode format as the default.

As someone who has developed their own implementation of `gitmoji-cli`, I still aim to maintain consistency and so have kept the shortcode format the default for my CLI tool as well. However, I think it may be now worth reviewing this decision and seeing if `emoji` makes more sense going forward. 

### Validations

- [X] Follow our [Code of Conduct](https://github.com/carloscuesta/gitmoji-cli/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/carloscuesta/gitmoji-cli/blob/master/.github/CONTRIBUTING.md).
- [X] Check that there isn't already an issue requesting the same feature.

mikelorant:
@carloscuesta Any thoughts on this proposal?

carloscuesta:
Hey!

Thanks for opening the issue 8 years ago already your comment made me feel old 🤣 
I think you have a valid point and still while some terminals will not support it people can always change back to the `:shortcode:` format so looks good to me!

mikelorant:
Pull request #1248 created making this change.

Unit tests pass successfully and deleting my global config and creating a new one with `gitmoji -g` showed the correct default set.

PR #1248, opened 2024-01-28
:sparkles: Set default emoji format to emoji

mikelorant:
The default emoji format was originally set to `code` for the following reason:

> The reason behind this was to be able to use the shortcode to favour
> the readability on certain terminals that do not support Emojis!
> @carloscuesta

Terminals are considerably more advanced on all platforms and the standard Gitmoji set is simple enough that all terminals would display them correctly nowadays.

This change updates the default format from `code` to `emoji`.

The benefits of this change:

- Improve display of emojis in commit titles, pull request titles and list of pull requests on some remote repository providers.

- Removes need to modify the pager to allow `git log` to display emojis.

Fixes: #1247


mikelorant:
@carloscuesta Can you approve this workflow? Thanks.

carloscuesta (approved):
LGTM!

Thanks 👏🏼

PR #1260, opened 2024-02-23
✨ Avoid text moving when nb char is 10

jdeniau:
## Description

When you type a commit message, on the 10nth char, the text move one character to the right because the prompt title goes from `9/48` to `10/48`

```diff
- Enter the commit title [9/48]: 123456789
+ Enter the commit title [10/48]: 1234567890
```

This is pretty annoying, so a added a `padStart(2)` to avoid this effect

```diff
- Enter the commit title [09/48]: 123456789
+ Enter the commit title [10/48]: 1234567890
```

There will still be the same effect after 99, but it's an un-recommended number of chars, so we might ignore this.

## Tests

- [ ] All tests passed.

### DIFF
--- a/src/commands/commit/prompts.js
+++ b/src/commands/commit/prompts.js
@@ -60,7 +60,9 @@ export default (
       message: 'Enter the commit title',
       validate: guard.title,
       transformer: (input: string) => {
-        return `[${(title || input).length}/${TITLE_MAX_LENGTH_COUNT}]: ${
+        const nbChar = String((title || input).length).padStart(2)
+        
+        return `[${nbChar}/${TITLE_MAX_LENGTH_COUNT}]: ${
           configurationVault.getCapitalizeTitle()
             ? capitalizeTitle(input)
             : input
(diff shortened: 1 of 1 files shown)


segersniels:
Visually this isn't the prettiest to look at. Wouldn't it make more sense to add a `0` in front?

src/commands/commit/prompts.js
```diff
@@ -60,7 +60,9 @@ export default (
       message: 'Enter the commit title',
       validate: guard.title,
       transformer: (input: string) => {
-        return `[${(title || input).length}/${TITLE_MAX_LENGTH_COUNT}]: ${
+        const nbChar = String((title || input).length).padStart(2)
+        
+        return `[${nbChar}/${TITLE_MAX_LENGTH_COUNT}]: ${
```
segersniels, line 65:
```suggestion
        const length = (title || input).length.toString().padStart(2, '0');
        
        return `[${length}/${TITLE_MAX_LENGTH_COUNT}]: ${
```

jdeniau:
@segersniels commited your suggestion

segersniels (approved):

Issue #1275, opened 2024-03-12
Warn the user if they are not in a git repo - bail out gracefully


mikelinhas:
### Description of the problem

When the tool is used outside a repo, it breaks and prints out the stdout from the caught error.

### Solution

Before running the create commit command (for example), first check if the user is in a git repository.

### Alternatives

This could be done before asking the user for input, since we know the tool will fail outside a repo.

### Additional context

I saw there is a Typescript migration proposal for this repo.
I programmed it with Flow, or at least I think I did, since I have no prior knowledge of Flow. 

### Validations

- [X] Follow our [Code of Conduct](https://github.com/carloscuesta/gitmoji-cli/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/carloscuesta/gitmoji-cli/blob/master/.github/CONTRIBUTING.md).
- [X] Check that there isn't already an issue requesting the same feature.

PR #1276, opened 2024-03-12
🚸 Warn user if they are not in a git repository

mikelinhas:
The change introduces a check to bail out if the user runs the tool outside a git repository. It will also warn the user. This handles this common pitfall gracefully instead of logging the stdout errors that happen in these cases.

## Description

Issue: #1275

## Tests

- [x] All tests passed.

Did not create a new test for this.


src/utils/isGitRepository.js
```diff
@@ -0,0 +1,13 @@
+// @flow
```
segersniels, line 1: We're in the middle of slowly moving away from Flow to TypeScript. So new files should definitely be written in TS.

src/commands/commit/withClient/index.js
```diff
@@ -5,9 +5,18 @@ import chalk from 'chalk'
 import isHookCreated from '@utils/isHookCreated'
 import configurationVault from '@utils/configurationVault'
 import { type Answers } from '../prompts'
+import isGitRepository from '@utils/isGitRepository'
 
 const withClient = async (answers: Answers): Promise<void> => {
   try {
+    if(!isGitRepository()) {
```
carloscuesta, line 12: I would probably avoid handling this through a helper and instead improve the way we show errors to users.

This way we can catch all exceptions handling them gracefully

src/commands/commit/withClient/index.js
```diff
@@ -5,9 +5,18 @@ import chalk from 'chalk'
...
+      return console.log(
+        chalk.red(
+          "\nError: Seems that you're trying to commit outside a git repository \n" +
+          'Please navigate to a git repository to use `gitmoji` \n'
+        )
+      )
+    }
```
carloscuesta, line 19:
```suggestion
```

carloscuesta: This is what's causing the error to display in a weird way, we can just remove this:

https://github.com/carloscuesta/gitmoji-cli/blob/68bc2a460f12f642fde5a0c3d5543fd055788b00/src/utils/isHookCreated.js#L16-L19

src/utils/isGitRepository.js
carloscuesta, line 1: We can completely get rid of this helper file 😊

carloscuesta (changes requested):
Thanks for the PR! 🙌🏼

src/utils/isGitRepository.js
mikelinhas: Thanks. I wasn't sure if I could just go ahead and write it in Typescript. 
And yes, it seems Carlos suggests removing the helper 😄

src/commands/commit/withClient/index.js
mikelinhas: The only issue I see is that it will not be a controlled error. 
We won't know inside the catch block what has actually happened.

Another thing to note - if we don't bail out with the helper, we ask the user for input knowing they are kind of doomed to fail. 🙈

carloscuesta: > Another thing to note - if we don't bail out with the helper, we ask the user for input knowing they are kind of doomed to fail. 🙈

Yep that's the downside of this approach, however I would say it should be a very unfrequent use-case and given that situation probably makes sense to do something that can improve the overall experience of the errors? 

> The only issue I see is that it will not be a controlled error.
> We won't know inside the catch block what has actually happened.

We will rely on `git` to tell the user what happened, I think this is better than wrapping specific use-cases only

segersniels: > We will rely on git to tell the user what happened, I think this is better than wrapping specific use-cases only
Yeah exactly my train of thought as well. I would just expose the actual errors to the user instead of masking them.

mikelinhas: Good point!

I reverted the changes with your recommendations. Now it breaks and the error is handled by the top level catch block.

That original error handling was nice, since it told you the git command it was going to execute before it failed. The problem now is that it fails in another git command and the message no longer makes sense. See the attached image for reference.

The only solution with this approach would be to make that error message more generic, which might be too generic 😞 

image

carloscuesta: Both are fine!

I don't think that loosing the `escapedCommand` thingy is too much of a loss, if we improve the overall error handling

So feel free to go with any of the options!

mikelinhas: Great! Thanks.
I removed the initial approach and just made the error handling more generic.
Hope it helps.

segersniels: Thanks for making the required changes! Ideally the files you touched are also converted to TS 🙏

mikelinhas: I'm a bit weary of migrating to Typescript file by file. 
I might be doing it wrong, but as soon as I change one file, I have to basically change all the depending files and tests in a cascading effect.
I'll keep an eye on the repo and lend a hand if I figure out how you're doing it.

carloscuesta (approved):
Thanks!!

Issue #1110, opened 2023-05-31
Transition away from Flow to TypeScript


segersniels:
### Discussion

Mostly just to open a discussion. 

I personally feel like TypeScript has been winning the race against Flow (and has won) for the last couple of years. So might be worth the effort to convert everything to TS and then compile/bundle the code into JS with something like `tsup`. Should improve development speed and give more control over types.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/carloscuesta/gitmoji-cli/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/carloscuesta/gitmoji-cli/blob/master/.github/CONTRIBUTING.md).
- [X] Check that there isn't already an issue requesting the same feature.

carloscuesta:
Hey!

This was one of the issues I wanted to open these days, a complete go from my side 🚀 
I believe there's a ton of improvement on how the codebase is typed (Flow got a bit outdated agree with this)

carloscuesta:
> > Is there an easy way we can make the migration incrementally? 🤔
> 
> 
> 
> We could set a rule that if you touch a file, you convert it to TS which will then gradually convert the codebase. Which is how I've done it in the past for large codebases. But might as well just do everything in one go since the codebase is not that big.
> 
> 
> 
> I had a go at it yesterday and shouldn't be too much work since it's purely types that change, the underlying javascript doesn't really change much. 

Sure but we would have to support both TS and JS Flow under the same build system, It's possible given that we can allow using JS within the `tsconfig` file and then transpile everything with tsup

Not sure if we will have conflicts by having "Flow types" on JS files when bundling with tsup

If that's the case then I would go for doing one PR with the setup for TS introducing the new bundling tool and then organise each part of the codebase so we can work together on a command/feature basis

segersniels:
Gave it another shot last night to see if I could get it working with both Flow and TS. Babel can be set up to support both typescript and flow, so that is pretty straight forward. Only thing I need to get working is the entire ESM flow and the fact that `import.meta` needs to be available.

Maybe we should look into transitioning away from `meow` first into something more robust so we can drop the `import.meta` and ESM requirement which kind of just complicates things over regular CJS imho.

carloscuesta:
@segersniels Great work!! I'll add the list of files that we need to migrate to your original message in a checklist mode!

Lauro235:
[...]

PR #1279, opened 2024-03-20
♻️ Maintenance: Improve TypeScript support

segersniels:
## Description

After reading reports from people trying their hand on converting files to TS, I noticed the setup wasn't fully allowing conversion without causing a big waterfall of required changes. While it was already possible to import Flow files into TS, it apparently wasn't yet possible to import TS files into Flow without the need for a lot of changes. This PR basically attempts to fix whatever was needed so proper TS work can start again 🚀

### Changed

- Adjusted `babel` to properly handle all `.ts` files and not just the ones in `./src`
- Adjusted `tsconfig.json` to properly output ESM so it can be used alongside other ESM modules like `meow` and `execa`
- `withClient.ts` has been converted to TS 

### Fixed

- Jest now properly looks for `.ts` files and includes them in the test run
- `cli.spec.ts` has not been running since the original TS introduction and is now running again

Issue: https://github.com/carloscuesta/gitmoji-cli/issues/1110#issuecomment-1996843360

## Tests

- [x] All tests passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,13 +92,18 @@
     "coverageDirectory": "./coverage/",
     "collectCoverage": true,
     "collectCoverageFrom": [
-      "src/**/*.js"
+      "src/**/*.js",
+      "src/**/*.ts"
     ],
     "testMatch": [
-      "**/*.(spec).(js)"
+      "**/*.(spec).(js|ts)"
     ],
     "setupFiles": [
       "./test/setupTests.js"
+    ],
+    "moduleFileExtensions": [
+      "js",
+      "ts"
     ],
     "transformIgnorePatterns": [],
     "moduleNameMapper": {
@@ -106,7 +111,8 @@
       "^#supports-color$": "<rootDir>/node_modules/chalk/source/vendor/supports-color/index.js",
       "@utils/(.*)$": "<rootDir>/src/utils/$1",
       "@commands/(.*)$": "<rootDir>/src/commands/$1",
-      "@constants/(.*)$": "<rootDir>/src/constants/$1"
+      "@constants/(.*)$": "<rootDir>/src/constants/$1",
+      "^(\\.\\.?\\/.+)\\.js$": "$1"
     }
   },
   "prettier": {
@@ -124,22 +130,14 @@
     ],
     "env": {
       "es2021": true,
-      "node": true
+      "node": true,
+      "jest": true
     },
     "extends": [
       "eslint:recommended",
       "plugin:@typescript-eslint/recommended"
     ],
     "overrides": [
-      {
-        "files": [
-          "./**/*.spec.js",
-          "./**/*.spec.ts"
-        ],
-        "env": {
-          "jest": true
-        }
-      },
       {
         "files": [
           "./**/*.ts"
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "target": "es6",
-    "module": "esnext",
-    "moduleResolution": "node",
+    "module": "node16",
+    "moduleResolution": "node16",
     "baseUrl": "./",
     "paths": {
       "@utils/*": ["src/utils/*"],
--- a/.flowconfig
+++ b/.flowconfig
@@ -15,6 +15,7 @@ module.name_mapper='^@utils\/\(.*\)$' -> '<PROJECT_ROOT>/src/utils/\1'
 module.name_mapper='^@constants\/\(.*\)$' -> '<PROJECT_ROOT>/src/constants/\1'
 module.name_mapper='^@commands\/\(.*\)$' -> '<PROJECT_ROOT>/src/commands/\1'
 exact_by_default=true
+module.file_ext=.ts
 
 [strict]
 
--- a/babel.config.json
+++ b/babel.config.json
@@ -22,7 +22,7 @@
   },
   "overrides": [
     {
-      "test": ["./src/**/*.ts"],
+      "test": ["**/*.ts"],
       "presets": ["@babel/preset-typescript"]
     }
   ]
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -3,10 +3,10 @@ import meow from 'meow'
 import updateNotifier from 'update-notifier'
 import { readFileSync } from 'fs'
 
-import FLAGS from '@constants/flags'
-import findGitmojiCommand from '@utils/findGitmojiCommand'
-import type { CommitOptions } from '@commands/commit'
-import type { SearchOptions } from '@commands/search'
+import FLAGS from '@constants/flags.js'
+import findGitmojiCommand from '@utils/findGitmojiCommand.js'
+import type { CommitOptions } from '@commands/commit/index.js'
+import type { SearchOptions } from '@commands/search/index.js'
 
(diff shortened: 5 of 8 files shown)


segersniels:
@carloscuesta I know the TS version of `cli.spec.js` hasn't been running since the refactor, but it seems that when I revert it back to the state it was before the refactor it doesn't even run. Did that test ever run? It seems like a bunch of stubs are missing to actually make everything work.

carloscuesta:
> @carloscuesta I know the TS version of `cli.spec.js` hasn't been running since the refactor, but it seems that when I revert it back to the state it was before the refactor it doesn't even run. Did that test ever run? It seems like a bunch of stubs are missing to actually make everything work. Not entirely familiar with jest as I usually work with mocha & chai so I might be missing something obvious.

Hey!

It's probably because of this:

https://github.com/carloscuesta/gitmoji-cli/blob/150a3505c9bd6d1c0012e7cc944b5e187a34a280/package.json#L97-L99

This is the pattern that jest uses to find which spec files to run, since the `cli.spec` was moved to `TS` we need to adjust this to:

```json
"testMatch": [
      "**/*.(spec).(js)",
      "**/*.(spec).(ts)"
    ],
```

So jest can find the `.spec.ts` files 👍🏼

segersniels:
> > @carloscuesta I know the TS version of `cli.spec.js` hasn't been running since the refactor, but it seems that when I revert it back to the state it was before the refactor it doesn't even run. Did that test ever run? It seems like a bunch of stubs are missing to actually make everything work. Not entirely familiar with jest as I usually work with mocha & chai so I might be missing something obvious.
> 
> Hey!
> 
> It's probably because of this:
> 
> https://github.com/carloscuesta/gitmoji-cli/blob/150a3505c9bd6d1c0012e7cc944b5e187a34a280/package.json#L97-L99
> 
> This is the pattern that jest uses to find which spec files to run, since the `cli.spec` was moved to `TS` we need to adjust this to:
> 
> ```json
> "testMatch": [
>       "**/*.(spec).(js)",
>       "**/*.(spec).(ts)"
>     ],
> ```
> 
> So jest can find the `.spec.ts` files 👍🏼

That was the original issue indeed, which we missed back when we refactored a couple of months ago. Test has been changed back to `cli.spec.js` in this PR and is running again, but failing on missing stubs. So I was wondering if it ever ran correctly or something broke with the jest mocks. You probably know more about jest than me

carloscuesta:
Yes, the test was running and passing successfully since it was added you can check it for example by going to the commit previous to the introduction of TS support and running the test:

```
git checkout c0b0ab8
yarn install
yarn test
```

Probably what happened is that as we missed including TS specs as part of that PR likely the test broke due to dependency updates that happened in the middle.

CleanShot 2024-03-21 at 22 08 37@2x

carloscuesta:
Actually the issue is that the mock was not properly implemented on that PR (not because any dependency update)

segersniels:
> Yes, the test was running and passing successfully from the beginning of it's inception you can check it for example by going to the commit previous to the introduction of TS support and running the test:
> 
> ```
> git checkout c0b0ab8
> yarn install
> yarn test
> ```
> 
> Actually the issue is that the mock was not properly implemented on that PR https://github.com/carloscuesta/gitmoji-cli/pull/1125/files#r1229243943
> 
> CleanShot 2024-03-21 at 22 08 37@2x

I reverted all changes for that file in this PR but it's still not running it seems. So it should be identical to what it was before the refactor.

package.json
```diff
@@ -61,6 +61,7 @@
   "devDependencies": {
     "@babel/cli": "7.23.9",
     "@babel/core": "7.24.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
```
segersniels, line 64: This one is apparently included in babel since `@babel/core@7.8.0`

carloscuesta:
It's tricky to exactly pin point what's the source of the issue now, as the spec file was not running since they refactor and several dependency updates were merged. What we certainly now is that before that specific commit it was passing as of this https://github.com/carloscuesta/gitmoji-cli/pull/1279#issuecomment-2013740537

So if you want to understand where the issue was exactly introduced (after reverting the TS PR) what I would suggest is doing a `bisect` https://git-scm.com/docs/git-bisect setting this hash as good `c0b0ab8` and the latest commit in master as bad

You'll have to revert the TS PR anyway bc the mock is not properly implemented

segersniels:
> > I reverted all changes for that file in this PR but it's still not running it seems. Which is weird to me because it should be identical to what it was before the refactor.
> 
> It's tricky to exactly pin point what's the source of the issue now, as the spec file was not running since the refactor and several dependency updates were merged. What we certainly know is that before that specific commit it was passing as of this [#1279 (comment)](https://github.com/carloscuesta/gitmoji-cli/pull/1279#issuecomment-2013740537)
> 
> So if you want to understand where the issue was exactly introduced (after reverting the TS PR) what I would suggest is doing a `bisect` https://git-scm.com/docs/git-bisect setting this hash as good [c0b0ab8](https://github.com/carloscuesta/gitmoji-cli/commit/c0b0ab8d9460400712718a49ff2191e4412c479a) and the latest commit in master as bad

My guess is the dynamic imports that we introduced as part of speed improvements.

tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "target": "es6",
-    "module": "esnext",
-    "moduleResolution": "node",
+    "module": "node16",
+    "moduleResolution": "node16",
```
carloscuesta, line 5: Why we need to change this? 🤔

package.json
```diff
@@ -92,21 +92,27 @@
...
+    ],
     "transformIgnorePatterns": [],
     "moduleNameMapper": {
       "^#ansi-styles$": "<rootDir>/node_modules/chalk/source/vendor/ansi-styles/index.js",
       "^#supports-color$": "<rootDir>/node_modules/chalk/source/vendor/supports-color/index.js",
       "@utils/(.*)$": "<rootDir>/src/utils/$1",
       "@commands/(.*)$": "<rootDir>/src/commands/$1",
-      "@constants/(.*)$": "<rootDir>/src/constants/$1"
+      "@constants/(.*)$": "<rootDir>/src/constants/$1",
+      "^(\\.\\.?\\/.+)\\.js$": "$1"
```
carloscuesta, line 115: This is to allow importing files without specifying the `.js` extension right?

tsconfig.json
segersniels: Else TS complains that it can't find the `execa` and `meow` modules. See https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

package.json
segersniels: This is required to make sure `jest` can resolve the new `.js` suffix imports. If we don't specify it the tests will fail because it can't find the modules.

tsconfig.json
carloscuesta: Thanks for the reference!

package.json
carloscuesta: I see, so part of the transition to make this an ESM (since it's a requirement to specify extension on ESMs)

carloscuesta (approved):

Issue #1412, opened 2025-03-27
Searching configuration


yzx9:
### Description of the problem

Searching for a configuration file beyond just the current working directory is useful in the following cases:
- Monorepo workflows: When working within a subpackage, but the configuration is located in the root directory.
- Global configuration: When a configuration file (e.g., `/home/user/.gitmojirc`) is placed at the user level, allowing it to be used from any subdirectory.
- Better support for the immutable configuration manager, e.g. Home-manager

### Solution

Current implementaion:

https://github.com/carloscuesta/gitmoji-cli/blob/142313b91eb0e7fb31314f4df1c8072f9d4fdc49/src/utils/configurationVault/getConfiguration.js#L60-L69

Solution:

```js
let currentDir = cwd()

while (currentDir !== dirname(currentDir)) {
  // put current implementation here

  currentDir = dirname(currentDir)
}
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/carloscuesta/gitmoji-cli/blob/master/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/carloscuesta/gitmoji-cli/blob/master/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue requesting the same feature.

PR #1422, opened 2025-04-10
✨ Searching config in ancestor directories

yzx9:
## Description

Searching for a configuration file beyond just the current working directory is useful in the following cases:
- Monorepo workflows: When working within a subpackage, but the configuration is located in the root directory.
- Global configuration: When a configuration file (e.g., `/home/user/.gitmojirc`) is placed at the user level, allowing it to be used from any subdirectory.
- Better support for the immutable configuration manager, e.g. Home-manager

close: #1412

## Tests

- [x] All tests passed.


carloscuesta (approved):
LGTM! 🚀 

Thanks for raising a PR

